PR #52, opened 2017-01-12
Push away from dome

### LINKED ISSUE #6: Improve magic dome
The magic dome implementation has some outstanding quirks:
-  [ ] The force field push throws the player slightly in the air. If outside the dome (or still on the border after the push) the vertical velocity will add, throwing the player higher and higher into the air. One way to check for this is to make sure that the impulse does _not_ cause the player to be further away from the center than before.
- [ ] Instead of pushing the player always back inside the dome it may be better to push away from the sphere (-> back inside if coming from inside, further away outside if trying to get in).
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36271395-improve-magic-dome?utm_campaign=plugin&utm_content=tracker%2F1483903&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1483903&utm_medium=issues&utm_source=github).
</bountysource-plugin>

krishnarb3:
Fixes #6  
The magic dome now pushes player outside if trying to come in and pushes back inside if trying to go out.
Also higher the altitude lesser the Y component of the impulse to prevent player from being pushed away too much.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
@@ -25,6 +25,7 @@ import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterSystem;
 import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
 import org.terasology.itemRendering.components.AnimateRotationComponent;
+import org.terasology.logic.actions.PlaySoundActionComponent;
 import org.terasology.logic.characters.CharacterImpulseEvent;
 import org.terasology.logic.characters.CharacterMoveInputEvent;
 import org.terasology.logic.location.LocationComponent;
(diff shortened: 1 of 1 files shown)


Cervator:
Tried testing this, but am having trouble finding my way outside. Attempted teleport a few times but either ended up in the wrong spot of squished from insta-impulsing around. Tried to fly out of the stop but also got stuck in some weird impulse pushing loop.

@krishnarb3 how did you go about testing it the best way?

@skaldarnar & @msteiger: poke! Sending this your way for further review, testing, and hopefully merging :-)

krishnarb3:
The issue with this is that teleport is seen as a movement, hence impulse acts as if the player moved from coordinate 1 to coordinate 2.
I tested by commenting out one part while testing the other (Eg: for testing impulse when player enters from outside, I commented out the impulse part for magic dome pushing player inside).
This can be solved by checking if the player is teleporting and disabling the magic dome at that moment alone.
Any suggestions on how to do that i.e. How to check if he is teleporting?

Cervator:
So looking real quick in `ServerCharacterPredictionSystem` method `onTeleport` handles the `CharacterReleportEvent` - possibly you could just add your own higher priority event handler somewhere in the Magic Dome System that could set a flag or take some other action to avoid the impulse?

skaldarnar:
This is scheduled on my TODO list once I've got a stable internet connection again. Can't wait to test that out 😄

krishnarb3:
Hello I am still stuck on disabling all other events if he/she is teleporting. Could someone help me with that?
Thanks.

skaldarnar:
Hi @krishnarb3, sorry for the long delay. I just managed to test this out, looks good to me. 

For testing, I shifted the dome away from player's spawn position so that I spawned outside. This is not ideal, but anyways. 

A possible solution to our problem might be not only check the player's before and after position, but also the velocity. By adding the following line to `onTeleport` the player's velocity will be canceled on teleport (thus, teleport will be a change of position without change of velocity):
```Java
newState.setVelocity(Vector3f.zero());
```
I'm not sure about other implications of this change, though.

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -77,17 +78,40 @@ public void onCharacterMovement(CharacterMoveInputEvent moveInputEvent, EntityRe
...
-                player.send(new CharacterImpulseEvent(impulse));
+        for (EntityRef domeEntity : entityManager.getEntitiesWith(MagicDome.class, LocationComponent.class)) {
+            LocationComponent domeLocationComponent = domeEntity.getComponent(LocationComponent.class);
+            Vector3f domeCenter = domeLocationComponent.getWorldPosition();
+            Vector3f domeVerticalTop = domeCenter.addY(WORLD_RADIUS);
 
-                player.send(new PlaySoundEvent(magicDomeEntity.getComponent(MagicDome.class).hitSound, 2f));
+            MagicDome dome = domeEntity.getComponent(MagicDome.class);
+
+            if (deltaDistance > 0.2f) {
```
skaldarnar: Here, we might check the player's velocity to catch a teleport event and don't act on it. Maybe you can give it a try?

Cervator:
@skaldarnar are you good with merging this since you approved? Just hit the button or push from local in that case, no need to wait :-)

Can always submit a follow-up issue if this is stable enough to merge then do a round two later.

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
krishnarb3: I'll try to do it asap

PR #53, opened 2017-07-05
update to new particle system - load particle prefab from resources

skaldarnar:
Requires https://github.com/Terasology/LightAndShadowResources/pull/19

I've added a TODO item for an outstanding issue regarding the particle texture. Again, might be an easy thing for @MaxBorsch.

Part of https://github.com/MovingBlocks/Terasology/issues/2941

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
@@ -21,6 +21,7 @@ import org.slf4j.LoggerFactory;
 import org.terasology.entitySystem.Component;
 import org.terasology.entitySystem.prefab.Prefab;
 import org.terasology.entitySystem.prefab.PrefabManager;
+import org.terasology.particles.components.ParticleEmitterComponent;
 import org.terasology.utilities.Assets;
 import org.terasology.audio.AudioManager;
 import org.terasology.entitySystem.entity.EntityManager;
@@ -95,7 +96,7 @@ public class CardSystem extends BaseComponentSystem {
             event.consume();
             return;
         }
-        logger.info("Prime block {} at pos {}", primeBlock, primePos);
+
         Block belowBlock = worldProvider.getBlock(primePos.x, primePos.y - 1, primePos.z);
         Block aboveBlock = worldProvider.getBlock(primePos.x, primePos.y + 1, primePos.z);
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
```diff
@@ -136,26 +137,11 @@ public void placeCard(ActivateEvent event, EntityRef entity) {
...
-        particleEffect.minLifespan = 1f;
-        particleEffect.maxLifespan = 1.5f;
-        particleEffect.targetVelocity.set(0, -5, 0);
-        particleEffect.acceleration.set(2f, 2f, 2f);
-        particleEffect.collideWithBlocks = true;
-        particlesEntity.addComponent(particleEffect);
+        //TODO change particle texture to match block texture
+        Prefab particleEffectPrefab = prefabManager.getPrefab("LightAndShadowResources:cardParticleEffect");
+        for (Component c : particleEffectPrefab.iterateComponents()) {
+            entity.addComponent(c);
```
skaldarnar: Right, most component should only be added once, despite the `ParticleEmitterComponent` itself. AFAIK it will just emit a fixed number of particles and then it will be destroyed after "dying". I'll look into this later.

skaldarnar:
Hm, seems like I need to revert some of the changes, i.e.,
- create a new entity for the particles (I first used a builder by hand, setting the location properly based on the card's location component)
- that entity will then be destroyed after the particle emission is finished, so no hassle cleaning that up
- this would also solve the problem of adding too many components to the card entity itself

Should be an easy fix ;-)

Cervator:
Merging updated version, thanks for the extra tweaks :-)

Works better now, could still provoke one problem that results in some noise in the log but no apparent functional issues - probably just some failure to clean up or trigger more particles:

```
22:05:37.741 [main] INFO  o.t.ligthandshadow.logic.CardSystem - current health: 1
22:05:38.124 [main] INFO  o.t.ligthandshadow.logic.CardSystem - current health: 0
22:05:38.125 [main] ERROR o.t.e.event.internal.EventSystemImpl - Failed to invoke event
java.lang.IllegalStateException: Entity EntityRef{id = 1693, netId = 0, prefab = 'LightAndShadowResources:spadesAceBlock'} is not a registered entity
	at com.google.common.base.Preconditions.checkState(Preconditions.java:199)
	at org.terasology.particles.updating.ParticleUpdaterImpl.dispose(ParticleUpdaterImpl.java:88)
	at org.terasology.particles.ParticleSystemManagerImpl.onEmitterDeactivated(ParticleSystemManagerImpl.java:91)
```

I also noted the icons in the toolbar went kinda funny, three of them ended up looking the same in the toolbar, but placing correctly as a particular suit. In any case L&S needs more work so this isn't much of a concern.

Thanks!

skaldarnar:
Uh, I must have messed it up when tweaking the prefabs... probably my regex search-and-replace was not as complete as I thought... I'll push an update for that directly to `master` 😉

PR #60, opened 2017-10-20
Update .gitignore

aaron-harris:
My Git was complaining about unversioned files in the `build` directory.  After poking around a bit, it looks like LightAndShadow has an older .gitignore that has since been replaced with a uniform file on most of the other modules (and that file ignores the `build` directory).  This PR is just to update LightAndShadow to use the newer file.


Cervator:
Thanks! And correct, L&S is a pretty early module, there are a few still on the old template :-)

Issue #43, opened 2016-10-25
"Gray" Flora Art Asset


rzats:
Idea to go with the "card suit color" flora blocks - an "unclaimed" version of the suit flowers that the player could tag with a tool of some sort to turn into flowers following their side, yielding some sort of benefit like territory claim / regeneration for nearby friendly.

Edit: There are existing faction-specific versions at https://github.com/Terasology/LightAndShadowResources/tree/master/assets/blockTiles/flora

rzats:
Art assets created by 
https://github.com/Terasology/LightAndShadowResources/pull/7 - only need to hook them up in-game now 🙂

Cervator:
Closing as completed by #62 !

Issue #12, opened 2016-10-25
Flora Generator


rzats:
A flora generator to use the custom L&S flora art assets requested in #11 is nice to have and should be a fairly simple modification to a similar system in the Core module, modified to use L&S-specific blocks.  See MovingBlocks/Terasology#544 for some legacy discussion on the subject.
##

rzats:
_From @Tuxacker on June 17, 2013 13:34_

I would make the Flora Generator more flexible by adding custom flower generation with a json file in the mod that describes all flora types which should be added and in which biomes / with which probability they should be added instead of cloning it. I could possibly do that, but i don't quite understand the mod data loading mechanics. Also this project is new to me, so i'm sorry in advance if I misunderstood something.

rzats:
_From @skaldarnar on June 17, 2013 17:20_

Pinging @immortius for mod loading mechanics. Also, make sure to check out the forums: 
http://forum.movingblocks.net

rzats:
_From @Cervator on June 17, 2013 17:52_

Talked to @Tuxacker some on IRC, idea boils down to the following rough outline:

1) Define a new asset type - [module]/worldgen/flora.json
2) Have engine scan for the new asset type
3) If found feed a list of plants from the flora.json to the FloraGenerator when running with that module active

@immortius - question then would be if you have any hints on how / where to do that or any links that might be helpful, or alternatively a better approach if you can think of any :-)

rzats:
_From @immortius on June 17, 2013 22:39_

I would suggest the following over an asset:
 
1. Using block categories.  Flora could be given a flora category and biome categories.  This is a little crude but is simple and open for mods adding new flora.

2. Add a Flora list component and then create an entity prefab or prefabs with the component. This saves creating a whole new asset type. Mods can introduce new lists.

3. Add a flora component and give each flora block a prefab with it.  This can be used to add whatever generation information is needed to the blocks.
This is a little awkward at the moment without inline prefab support in block definitions.

While I've started working towards custom asset support in mods it isn't ready yet, and even then I'd suggest an entity system aligned approach where possible.

Cervator:
Closing as completed by #62 !

Issue #11, opened 2016-10-25
Flora Art Assets


rzats:
Rather than the standard flowers currently present in the game we could create custom flavourful flowers representing the [4 different suits of cards](https://en.wikipedia.org/wiki/Suit_%28cards%29) as flowers. This might also include some appropriately colored grass types.
##

dacharya64:
Looking at the current state of the module, it looks like the Diamonds and Clubs flora assets still need to be done. Is that correct? If so I can work on that.

Cervator:
I believe so, yes. You may also want to go through the L&S thread in the art forum. It is pretty long - lots of forgotten stuff in there :-)

Also issues/PRs over in LightAndShadowResources just in case

dacharya64:
Made a PR for this at https://github.com/Terasology/LightAndShadowResources/pull/22

Cervator:
Closing this as completed by https://github.com/Terasology/LightAndShadowResources/pull/22 - thanks @dacharya64 ! 

While we could still add grasses and other things we've got a pretty decent set now and some other artwork requesting issues so it seems fair to close this with the expectation to see more like it in the future :-)

PR #62, opened 2018-03-03
Issue #12 Adds a rasterizer to place L&S flora in world on world generation

dacharya64:
PR for Issue #12 -- adds a FloraGenerator for L&S-specific assets (listed in #11 and #43) and calls it on world generation, placing L&S flora throughout the world.

NOTE: Although this works now, the code is a bit redundant and I wasn't sure how to reduce redundancy while still getting the flora to spawn correctly, so any notes on that would be helpful

### DIFF
--- a/src/main/java/org/terasology/las/LaSFloraRasterizer.java
+++ b/src/main/java/org/terasology/las/LaSFloraRasterizer.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


dacharya64:
Yeah, that's what I was thinking about--reusing the gen code from Cities (or in general, I suppose) in a way that you can swap different rasterizers in/out without redeclaring everything in the worldbuilder in a new generator

Cervator:
World plugins make up the best solution we have there so far, but I'm not sure how far they can reach when you're really overhauling a world. Simply placing flora is definitely a candidate for a plugin - see https://github.com/Terasology/AlchemyPlantGenerator for an example. But L&S goes a way past just that :-)

PR #63, opened 2018-03-07
"Dome" console command to toggle dome 

dacharya64:
This allows the player to toggle the magic barrier (dome) on and off using the console command "dome". If the barrier is up, it is removed on running the command, and vice versa.


Cervator:
Am presently merging many things, thought I'd get this out of the way too to help clean up in Light & Shadow land :-) Even if `master` and `develop` have diverged a bit more. I'm sure we can sort that out sometime.

PR #65, opened 2018-05-30
Adds faction component to player, changes flag pickup logic, and starts player on platform

dacharya64:
Players now have faction (by default it is white (neutral) and now spawn at a platform above the capture the flag play area (and can see down below). They can only pick up the flag by pressing `E` now if they are of the opposite team (only red players can pick up black flags, and vice versa. White players cannot pick up either flag). 

Incoming L&S Resources PR will fix the issue with players destroying the flag block if they are on the same team.

PR #66, opened 2018-06-05
Merge diverged develop back into master

Cervator:
This might just be a throwaway PR considering #65 exists and @dacharya64 might be a better person to do this. I rushed conflict resolution a bit too and at first still had the behavior test tools in.

I wanted to see how relatively straight forward this merge would be and it *seems* like this could be done just about any time now.

Let me know what you think @dacharya64 :-)


Cervator:
Oh this would go with https://github.com/Terasology/Cities/pull/34 as well

dacharya64:
Seems like it would work to me too!

PR #67, opened 2018-06-13
Adding beginning elements of score system and UI

dacharya64:
HUD for score should now display at the top of the screen. Does not yet update score based on player actions (that's next on my list).

PR #68, opened 2018-06-24
Player vs Player Combat Mechanics (WIP)

dacharya64:
This PR encompasses the following changes: 
-Giving magic staff to player on teleport (complete) (Test with https://github.com/Terasology/LightAndShadowResources/pull/26)
-Shooting raycast out from staff on activate
-Checking to see if the raycast hits someone
-If raycast does hit someone and that someone is holding a flag, the flag is dropped


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -0,0 +1,104 @@
...
+    @ReceiveEvent(components = {WinConditionCheckOnActivateComponent.class, LASTeamComponent.class})
+    public void onActivate(ActivateEvent event, EntityRef entity) {
+        LASTeamComponent baseTeamComponent = entity.getComponent(LASTeamComponent.class);
+        EntityRef player = event.getInstigator();
+        CharacterHeldItemComponent characterHeldItemComponent = player.getComponent(CharacterHeldItemComponent.class);
+        if (characterHeldItemComponent != null) {
+            EntityRef heldItem = characterHeldItemComponent.selectedItem;
+
+            //check to see if player has other team's flag
+            if (baseTeamComponent.team.equals(baseTeamComponent.RED) && heldItem.getComponent(BlockItemComponent.class).blockFamily.getURI().toString()
```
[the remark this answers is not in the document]
vampcat: Also, regarding the if : You have split it at a very wrong point. It legit took me 10 seconds to figure out that the second expression is continued on the second line - which is a lot.

Split the condition on &&, and convert `LightAndShadowResources:BlackFlag` and `LightAndShadowResources:redFlag` to their own constants. Make the second condition more readable if you can.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TakeBlockOnActivationSystem.java
```diff
@@ -58,11 +58,11 @@ public void onActivate(ActivateEvent event, EntityRef entity) {
...
 
         BlockComponent blockComponent = entity.getComponent(BlockComponent.class);
-        LASTeam flagTeamComponent = entity.getComponent(LASTeam.class);
+        LASTeamComponent flagTeamComponent = entity.getComponent(LASTeamComponent.class);
 
 
         EntityRef flagTaker = event.getInstigator();
-        LASTeam playerTeamComponent = flagTaker.getComponent(LASTeam.class);
+        LASTeamComponent playerTeamComponent = flagTaker.getComponent(LASTeamComponent.class);
         //if the flag being taken is a red flag and the player is on the black team, let them take the flag
```
vampcat: //if -> // If

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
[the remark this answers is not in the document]
vampcat: Also, what is the need for `equalsIgnoreCase`? `equals` should be more than adequate for your purposes.

Here is the reason why: You've used `LightAndShadowResources:redFlag` at some point in this PR, make the entire system wonky.

There is a reason I insist on consistency across everything - to avoid things like this.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -0,0 +1,104 @@
...
+    private int blackScore = 0;
+    private UILabel blackScoreArea;
+    private UILabel redScoreArea;
+
+    @Override
+    public void postBegin() {
+        // Sets score screen bindings
+        scoreScreen = nuiManager.getHUD().getHUDElement("LightAndShadow:ScoreHud");
+        blackScoreArea = scoreScreen.find("blackScoreArea", UILabel.class);
+        blackScoreArea.bindText(new ReadOnlyBinding<String>() {
```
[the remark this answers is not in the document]
dacharya64: I'm not super familiar with how lambdas work, is there an example of this in Terasology I could look at?

skaldarnar:
It looks like this PR is based on #67 which makes it hard to review (review only the relevant changes for this PR, that is). If you work on different parts of the code that are unrelated to each other you should branch off from `master` instead of another feature branch.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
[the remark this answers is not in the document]
skaldarnar: I'm not sure lambda can be applied here(?) They are syntactic sugar for an implementation of an interface with exactly one method (i.e., you only have to specify the implementation of this one method). Usually IntelliJ will also suggest to use a Lambda whenever possible...

To find usages of a specific class or interface you should also make use of the capablilites of your IDE :wink:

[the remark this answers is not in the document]
vampcat: As skaldy said, lambdas are just syntactic sugar. Just google around, the first couple of links all look promising.

Basically you can replace
```
        blackScoreArea.bindText(new ReadOnlyBinding<String>() {
            @Override
            public String get() {
                return String.valueOf(blackScore);
            }
        });
```
with
```
blackScoreArea.bindText(() -> String.valueOf(blackScore));
```
More compact, hence easier to understand.

@skaldarnar I don't see why we can't use lambda here. Am I missing something?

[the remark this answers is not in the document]
dacharya64: Cool, I've seen lambdas before but wasn't sure if you were referring to some special use in Terasology or not.

dacharya64:
This PR now encompasses all of the changes for the scoring and raycasting code.

dacharya64:
With the latest push, user should now be able to press the activate button (E) when targeting another player while holding the magic staff, and if that player is holding a flag, that flag is dropped.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -0,0 +1,103 @@
...
+import org.terasology.rendering.nui.NUIManager;
+import org.terasology.rendering.nui.databinding.ReadOnlyBinding;
+import org.terasology.rendering.nui.widgets.UILabel;
+import org.terasology.world.block.items.BlockItemComponent;
+
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class ScoreSystem extends BaseComponentSystem {
+    private static final Logger logger = LoggerFactory.getLogger(ScoreSystem.class);
+    private static final String BLACK_FLAG_URI = "LightAndShadowResources:blackFlag";
+    private static final String RED_FLAG_URI = "LightAndShadowResources:redFlag";
```
skaldarnar: You are defining the same constants in multiple locations (ScoreSystem, AttackSystem). Ideally, it should be statically available from done utility context.

dacharya64: I know this is being done wrong at the moment (constants should only be defined in one place) but wasn't sure where that one place should be. What do you mean by "done utility context"? I was thinking of defining them in the `FlagComponent` file and accessing from there, would that make sense?

skaldarnar: Should have read "some utility". Not sure whether components should have constants specific to instances kind of but it's a start

nihal111: Nah, shouldn't be in the component. @dacharya64 what @skaldarnar meant was to have a Utility class outside that's probably package private and holds the strings as `public static final` making them accessible to multiple other classes from the same package.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -0,0 +1,86 @@
...
+
+            // If raycast hits another player and another player has flag, make player drop flag
+            if (entity.hasComponent(PlayerCharacterComponent.class) && entity.hasComponent(CharacterHeldItemComponent.class)) {
+                CharacterHeldItemComponent characterHeldItemComponent = entity.getComponent(CharacterHeldItemComponent.class);
+                EntityRef heldItem = characterHeldItemComponent.selectedItem;
+                if (heldItem.getComponent(BlockItemComponent.class).blockFamily.getURI().toString().equals(BLACK_FLAG_URI)
+                        || heldItem.getComponent(BlockItemComponent.class).blockFamily.getURI().toString().equals(RED_FLAG_URI)) {
+                    Vector3f position = new Vector3f(attackingPlayer.getComponent(LocationComponent.class).getLocalPosition());
+                    Vector3f direction = localPlayer.getViewDirection();
+                    Vector3f newPosition = new Vector3f(position.x + direction.x * 1.5f,
```
nihal111: How much is 1.5f?
We should have this value as `static final` above.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/EnemySystem.java
```diff
@@ -53,7 +53,7 @@ public void initialise() {
     }
 
     @ReceiveEvent
-    public void onActivated(OnActivatedComponent event, EntityRef entityRef, LASTeam team, LocationComponent locationComponent) {
+    public void onActivated(OnActivatedComponent event, EntityRef entityRef, LASTeamComponent team, LocationComponent locationComponent) {
         logger.info("activated " + entityRef + " " + team.team);
```
nihal111: Please remove log statements when this PR is merge ready

Cervator:
Did another round of testing and suggestions. Going to merge this now to make testing easier and a next PR cleaner :-)

PR #69, opened 2018-07-02
Improvements to Capture the Flag attack system [WIP]

dacharya64:
Improvements for the capture the flag gameplay mode. 

Currently is work in progress for having the AttackSystem make the player drop a flag if it's in the player's inventory, not just being held.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/components/HasFlagComponent.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/components/HasFlagComponent.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.ligthandshadow.componentsystem.components;
+
+import org.terasology.entitySystem.Component;
+
+public class HasFlagComponent implements Component {
+    public String flag;
+}
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/components/HasFlagComponent.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/components/HasFlagComponent.java
@@ -19,4 +19,11 @@ import org.terasology.entitySystem.Component;
 
 public class HasFlagComponent implements Component {
     public String flag;
+
+    public HasFlagComponent() {
+    }
+
+    public HasFlagComponent(String flag) {
+        this.flag = flag;
+    }
 }
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
@@ -62,6 +62,7 @@ public class AttackSystem extends BaseComponentSystem {
     private BlockManager blockManager;
 
     private EntityRef flagSlot;
+    private EntityRef item;
 
 
     @ReceiveEvent(components = {FlagDropOnActivateComponent.class, PlayerCharacterComponent.class, HasFlagComponent.class})
@@ -117,16 +118,33 @@ public class AttackSystem extends BaseComponentSystem {
     @ReceiveEvent(components = {LASTeamComponent.class})
     public void onInventorySlotChanged(InventorySlotChangedEvent event, EntityRef entity) {
         EntityRef playerEntity = entity;
-        EntityRef item = event.getNewItem();
-        if (item.hasComponent(BlackFlagComponent.class)) { //&& playerEntity.getComponent(LASTeamComponent.class).team.equals(LASUtils.BLACK_TEAM)
-            worldProvider.setBlock(new Vector3i(LASUtils.CENTER_BLACK_BASE_POSITION.x, LASUtils.CENTER_BLACK_BASE_POSITION.y + 1, LASUtils.CENTER_BLACK_BASE_POSITION.z), blockManager.getBlock(LASUtils.BLACK_FLAG_URI));
-            inventoryManager.removeItem(playerEntity, EntityRef.NULL, item, true, 1);
-            return;
+        moveFlagToBaseIfPickedBySameTeamPlayer(event, playerEntity);
+    }
+
+    private void moveFlagToBaseIfPickedBySameTeamPlayer(InventorySlotChangedEvent event, EntityRef playerEntity) {
+        String flagTeam = checkWhichFlagPicked(event);
+        if (flagTeam != null) {
+            if (isPlayerInTeam(flagTeam, playerEntity)) {
+                worldProvider.setBlock(LASUtils.getFlagLocation(flagTeam), blockManager.getBlock(LASUtils.getFlagURI(flagTeam)));
+                inventoryManager.removeItem(playerEntity, EntityRef.NULL, item, true, 1);
+            }
         }
-        if (item.hasComponent(RedFlagComponent.class) && playerEntity.getComponent(LASTeamComponent.class).team.equals(LASUtils.RED_TEAM)) {
-            worldProvider.setBlock(new Vector3i(LASUtils.CENTER_RED_BASE_POSITION.x, LASUtils.CENTER_RED_BASE_POSITION.y + 1, LASUtils.CENTER_RED_BASE_POSITION.z), blockManager.getBlock(LASUtils.RED_FLAG_URI));
-            inventoryManager.removeItem(playerEntity, EntityRef.NULL, item, true, 1);
-            return;
+    }
+
+    private boolean isPlayerInTeam(String flagTeam, EntityRef playerEntity) {
+        return playerEntity.getComponent(LASTeamComponent.class).team.equals(flagTeam);
+    }
+
+    private String checkWhichFlagPicked(InventorySlotChangedEvent event) {
+        item = event.getNewItem();
+        if (item.hasComponent(BlockItemComponent.class)) {
+            if (item.hasComponent(BlackFlagComponent.class)) {
+                return LASUtils.BLACK_TEAM;
+            }
+            if (item.hasComponent(RedFlagComponent.class)) {
+                return LASUtils.RED_TEAM;
+            }
         }
+        return null;
     }
 }
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
@@ -31,19 +31,22 @@ import org.terasology.logic.characters.CharacterHeldItemComponent;
 import org.terasology.logic.common.ActivateEvent;
 import org.terasology.logic.inventory.InventoryManager;
 import org.terasology.logic.inventory.events.DropItemRequest;
+import org.terasology.logic.inventory.events.GiveItemEvent;
+import org.terasology.logic.inventory.events.InventorySlotChangedEvent;
 import org.terasology.logic.location.LocationComponent;
 import org.terasology.logic.players.LocalPlayer;
 import org.terasology.logic.players.PlayerCharacterComponent;
 import org.terasology.math.geom.Vector3f;
 import org.terasology.math.geom.Vector3i;
 import org.terasology.physics.Physics;
+import org.terasology.protobuf.EntityData;
 import org.terasology.registry.In;
 import org.terasology.world.WorldProvider;
 import org.terasology.world.block.BlockManager;
 import org.terasology.world.block.items.BlockItemComponent;
 
 @RegisterSystem(RegisterMode.AUTHORITY)
-public class AttackSystem extends BaseComponentSystem implements UpdateSubscriberSystem {
+public class AttackSystem extends BaseComponentSystem {
     @In
     InventoryManager inventoryManager;
     @In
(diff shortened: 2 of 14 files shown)


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -76,4 +85,39 @@ public void onActivate(ActivateEvent event, EntityRef entity) {
             }
         }
     }
+
+    @Override
+    public void update(float delta) {
```
jellysnake: This is a fairly intensive check to run every single frame. Can we listen to the `InventorySlotChangedEvent` instead and then check if the new item is a flag?
Then if it is we handle the team cases

dacharya64:
Things that should be working as of latest push: 
1) Flag position reset (sent back to base) on score increase
2) When player picks up flag of same team as they are, flag is removed from inventory and set back at base
Latest push changes the update() method for point 2 to onInventorySlotChanged() but I'm not 100% certain if this is works across different serves / clients -- was doing some testing and it didn't always remove the item from the client's inventory, I think this might have to do with using `private LocalPlayer localPlayer;` for the cleck? Would appreciate if someone else could test this out on multiplayer

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -22,66 +22,108 @@
...
+
+    // Checks to see if player picks up flag of the same team
+    // If so, moves flag back to base
+    @ReceiveEvent
+    public void onInventorySlotChanged(InventorySlotChangedEvent event, EntityRef entity) {
+        EntityRef playerEntity = entity;
+        EntityRef item = event.getNewItem();
+        if (playerEntity.hasComponent(LASTeamComponent.class) && item.hasComponent(BlockItemComponent.class)) {
+            // If player team and flag team are the same, return flag to base
+            if (item.getComponent(BlockItemComponent.class).blockFamily.getURI().toString().equals(LASUtils.RED_FLAG_URI)
```
nihal111: We should stop doing URI matches for checks.
Let's create a `RedFlagComponent`. Define the flag prefabs to have this component by default. All we need to do is check if the item has this component or not from now.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -22,66 +22,108 @@
...
+                }
+            }
+        }
+    }
+
+    // Checks to see if player picks up flag of the same team
+    // If so, moves flag back to base
+    @ReceiveEvent
+    public void onInventorySlotChanged(InventorySlotChangedEvent event, EntityRef entity) {
+        EntityRef playerEntity = entity;
```
nihal111: I think you could simply have the method define the EntityRef as `playerEntity` initially, but it'd be better to create a separate methods for this too-

0. Create a method that has all of the logic here, call it- `moveFlagToBaseIfPickedBySameTeamPlayer()` (or a better shorter version if you come up with any.
1. `private String checkWhichFlagPicked(event)`: This gets the new picked item, checks if it has BlockItemComponent, and if it does,
    - it checks if it has the RedFlagComponent (returns `LASUtils.RED_TEAM`)
    - it checks if it has the BlackFlagComponent (returns `LASUtils.BLACK_TEAM`)
    - else (returns `LASUtils.RED_TEAM`)
2. `private boolean isPlayerInTeam(String team)`: Check if player is on the same team as returned team from above method.
3. Inside LASUtils create a method `private Vector3i getFlagLocation(String team)` that returns the flag location as per the team passed.

Everything combined would look like this inside `moveFlagToBaseIfPickedBySameTeamPlayer`
```
String flagTeam = checkWhichFlagPicked(event);
if (isPlayerInTeam(flagTeam) {
    worldProvider.setBlock(LASUtils.getFlagLocation(team), blockManager.getBlock(LASUtils.RED_FLAG_URI))
    inventoryManager.removeItem(playerEntity, EntityRef.NULL, item, true, 1);
}
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -22,66 +22,108 @@
...
+        EntityRef item = event.getNewItem();
+        if (playerEntity.hasComponent(LASTeamComponent.class) && item.hasComponent(BlockItemComponent.class)) {
+            // If player team and flag team are the same, return flag to base
+            if (item.getComponent(BlockItemComponent.class).blockFamily.getURI().toString().equals(LASUtils.RED_FLAG_URI)
+                    && playerEntity.getComponent(LASTeamComponent.class).team.equals(LASUtils.RED_TEAM)) {
+                worldProvider.setBlock(new Vector3i(LASUtils.CENTER_RED_BASE_POSITION.x, LASUtils.CENTER_RED_BASE_POSITION.y + 1, LASUtils.CENTER_RED_BASE_POSITION.z), blockManager.getBlock(LASUtils.RED_FLAG_URI));
+                inventoryManager.removeItem(playerEntity, EntityRef.NULL, item, true, 1);
+                return;
+            }
+            if (item.getComponent(BlockItemComponent.class).blockFamily.getURI().toString().equals(LASUtils.BLACK_FLAG_URI)
```
nihal111: Create `BlackFlagComponent` for check

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -22,66 +22,108 @@
...
-                    entity.send(new DropItemRequest(heldItem, entity, impulseVector, newPosition));
+            if (entity.hasComponent(PlayerCharacterComponent.class)) {
+                blackFlagSlot = searchInventoryForFlag(entity, LASUtils.BLACK_FLAG_URI);
+                if (blackFlagSlot != EntityRef.NULL) {
+                    Vector3f position = new Vector3f(entity.getComponent(LocationComponent.class).getLocalPosition());
+                    Vector3f impulseVector = new Vector3f(attackingPlayer.getComponent(LocationComponent.class).getLocalPosition());
+                    entity.send(new DropItemRequest(blackFlagSlot, entity, impulseVector, position));
+                    return;
+                }
+                redFlagSlot = searchInventoryForFlag(entity, LASUtils.RED_FLAG_URI);
```
nihal111: Duplicate code below, NOOOOOOO

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -89,13 +98,19 @@ public void onActivate(ActivateEvent event, EntityRef entity) {
```
nihal111: Let's move the entire logic here inside `onActivate` to a method called- `checkAndResetGameOnScore`

This would do three things, have one method for each- 
1. Check if any team did score. (Check for the team in base's LASTeamComponent and for the heldItem to have the opposite team's FlagComponent)
2. Increment score. 
3. Reset the round- (This should be the same method for both teams and should completely reset the round however we want)

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -89,13 +98,19 @@ public void onActivate(ActivateEvent event, EntityRef entity) {
...
+                worldProvider.setBlock(new Vector3i(LASUtils.CENTER_BLACK_BASE_POSITION.x, LASUtils.CENTER_BLACK_BASE_POSITION.y + 1, LASUtils.CENTER_BLACK_BASE_POSITION.z), blockManager.getBlock(LASUtils.BLACK_FLAG_URI));
             }
 
             if (baseTeamComponent.team.equals(LASUtils.BLACK_TEAM)
+                    && heldItem.hasComponent(BlockItemComponent.class)
                     && heldItem.getComponent(BlockItemComponent.class).blockFamily.getURI().toString().equals(LASUtils.RED_FLAG_URI)) {
                 blackScore++;
+                inventoryManager.removeItem(player, player, heldItem, true);
+                worldProvider.setBlock(new Vector3i(LASUtils.CENTER_RED_BASE_POSITION.x, LASUtils.CENTER_RED_BASE_POSITION.y + 1, LASUtils.CENTER_RED_BASE_POSITION.z), blockManager.getBlock(LASUtils.RED_FLAG_URI));
             }
```
nihal111: I don’t see where the flag is removed from the losing player’s inventory and reset to it’s base (the losing player might also be halfway through running with the other player’s flag)

Have you not implemented this or am I missing something?

dacharya64: Correct, this hasn't been implemented yet

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
dacharya64: So I've had a problem from the beginning with finding a `Flag` or `BlackFlag` / `RedFlag` component attached to a flag block. In running `dumpEntities` I find an entity with the BlackFlagComponent attached:

```
{
      "id": 71,
      "parentPrefab": "engine:blockType",
      "engine:BlockType": {
        "block": "LightAndShadowResources:blackFlag"
      },
      "LightAndShadow:Flag": {},
      "LightAndShadow:LASTeam": {
        "team": "black"
      },
      "LightAndShadow:BlackFlag": {},
      "LightAndShadow:TakeBlockOnActivate": {}
    },
```

but in the inventory is a different entity reference: 
```
"entity": [
    {
      "id": 2321,
      "parentPrefab": "engine:blockItemBase",
      "owner": 132,
      "engine:DisplayName": {
        "name": "Blackflag"
      },
      "engine:BlockItem": {
        "blockFamily": "LightAndShadowResources:blackFlag"
      },
      "engine:BoxShape": {
        "extents": [
          0.92,
          0.92,
          0.92
        ]
      },
      "engine:Item": {
        "stackId": "block:LightAndShadowResources:blackFlag"
      },
      "engine:Mesh": {
        "translucent": false,
        "material": "engine:terrain",
        "color": [
          255,
          255,
          255,
          255
        ],
        "selfLuminance": 0.0,
        "hideFromOwner": false,
        "mesh": "engine:blockmesh#LightAndShadowResources:blackFlag"
      }
    },
```
This is why checking if the blockFamily String = URI works, but checking for the prefab does not. But I'm not sure why this is the case, and how to get a check for the prefab to work properly. Right now checking if the flag has a component, even when it does, fails.

I think this could be because it's a block that's placed into the world in the world render? But other components work, like the `TakeBlockOnActivate` check.

dacharya64:
As of the latest push, all code should now be working (ScoreSystem and AttackSystem functioning as needed) and the above suggestions all made for code refactoring. Also includes a system for detecting when either team has reached some max score and can reset level accordingly.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -22,66 +22,129 @@
...
+    @ReceiveEvent(components = {FlagDropOnActivateComponent.class, PlayerCharacterComponent.class, HasFlagComponent.class})
     public void onActivate(ActivateEvent event, EntityRef entity) {
-        // Entity refers to the thing being activated (in this case the other player)
-        EntityRef attackingPlayer = event.getInstigator(); // The player using the staff
+        dropFlagOnPlayerAttack(event, entity);
+    }
 
-        // If the attacking player is holding the magic staff when activating
-        if (attackingPlayer.getComponent(CharacterHeldItemComponent.class).selectedItem.hasComponent(RaycastOnActivateComponent.class)) {
+    /** When player activates another with the magic staff, checks to see if the attacked player has a flag
```
nihal111: new line after `/**`

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -22,66 +22,129 @@
...
+     * If so, makes the player drop the flag
+     * targetPlayer is player being attacked */
```
nihal111: newline before `*/`

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -22,66 +22,129 @@
...
+    private boolean canPlayerAttack(EntityRef attackingPlayer) {
+        if (!attackingPlayer.hasComponent(CharacterHeldItemComponent.class)) {
+            return false;
+        }
+        EntityRef heldItem = attackingPlayer.getComponent(CharacterHeldItemComponent.class).selectedItem;
+        return heldItem.hasComponent(RaycastOnActivateComponent.class);
+    }
+
+    // Checks to see if player picks up flag of the same team
+    // If so, moves flag back to base
```
nihal111: change comments to javadoc

src/main/java/org/terasology/ligthandshadow/componentsystem/components/HasFlagComponent.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.ligthandshadow.componentsystem.components;
+
+import org.terasology.entitySystem.Component;
+
+public class HasFlagComponent implements Component {
```
nihal111: can you add javadoc for this component, explaining which entities can have this component

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -24,34 +24,41 @@
...
 import org.terasology.rendering.nui.databinding.ReadOnlyBinding;
 import org.terasology.rendering.nui.widgets.UILabel;
+import org.terasology.world.WorldProvider;
+import org.terasology.world.block.BlockManager;
 import org.terasology.world.block.items.BlockItemComponent;
 
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class ScoreSystem extends BaseComponentSystem {
     private static final Logger logger = LoggerFactory.getLogger(ScoreSystem.class);
+    private static final int GOAL_SCORE = 5;
```
nihal111: Noooooooooooo, this goes in LASUtils 😅

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TakeBlockOnActivationSystem.java
```diff
@@ -61,16 +62,17 @@ public void onActivate(ActivateEvent event, EntityRef entity) {
...
+            flagTaker.addComponent(new HasFlagComponent(LASUtils.RED_TEAM));
             inventoryManager.giveItem(flagTaker, EntityRef.NULL, blockFactory.newInstance(blockManager.getBlockFamily(LASUtils.RED_FLAG_URI)));
             worldProvider.setBlock(blockComponent.getPosition(), blockManager.getBlock(BlockManager.AIR_ID));
             entity.destroy();
         }
         if (flagTeamComponent.team.equals(LASUtils.BLACK_TEAM) && playerTeamComponent.team.equals(LASUtils.RED_TEAM)) {
+            flagTaker.addComponent(new HasFlagComponent(LASUtils.BLACK_TEAM));
             inventoryManager.giveItem(flagTaker, EntityRef.NULL, blockFactory.newInstance(blockManager.getBlockFamily(LASUtils.BLACK_FLAG_URI)));
             worldProvider.setBlock(blockComponent.getPosition(), blockManager.getBlock(BlockManager.AIR_ID));
             entity.destroy();
```
nihal111: I know the above is not strictly a part of this PR, but I'd love it if you added a different method inside the onActivate with an intuitive name.
Also the above if condition has duplicate code, so maybe it can be extracted to a separate method that gets called with Black/Red team.

Cervator:
NIce work! Although it looks like you accidentally put a WIP commit on top of this branch for the scoring hud visibility in multiplayer that broke stuff. I was able to pull this PR minus that one commit and it works great, both sides can score in multiplayer, although only a server/client game instance can *see* said score :-)

I'm going to close this as merged so we can begin a nice clean PR for the next round of changes. The PR will say closed since I'm not including the accidental commit, but it really is merged!

Cervator:
See the network graph for the dangling commit that remains: https://github.com/Terasology/LightAndShadow/network

(although that'll change soon enough - right now it makes sense)

PR #73, opened 2018-08-11
[WIP] HUD now reskins with red hearts/black spades on team selection 

dacharya64:
Player health HUD now changes symbols based on the team players pick. Default is rounded circles (white), with the symbols of the health HUD changing to black spades and/or red hearts on team selection.

This is a work in progress because of a bug causing the change in NUIManager to change the HUD across all players rather than just the player making a team selection.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
@@ -17,14 +17,13 @@ package org.terasology.ligthandshadow.componentsystem;
 
 import org.terasology.math.geom.Vector3f;
 import org.terasology.math.geom.Vector3i;
-import org.terasology.registry.In;
-import org.terasology.world.block.BlockManager;
 
 public final class LASUtils {
     public static final String BLACK_FLAG_URI = "lightAndShadowResources:blackFlag";
     public static final String RED_FLAG_URI = "lightAndShadowResources:redFlag";
     public static final String RED_TEAM = "red";
     public static final String BLACK_TEAM = "black";
+    public static final String WHITE_TEAM = "white";
     /**
      * Position of Red base
      */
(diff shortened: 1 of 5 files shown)


Cervator:
Can confirm that it works! Looks nice :-)

As for the bug: Is that just because `LASSystem` is current plainly decorated with `@RegisterSystem` which means it defaults to `ALWAYS` - aka it'll run on *both* authority and client?

Also, you've got some conflicts in this branch vs master, might want to try a quick rebase :-)

dacharya64:
So still not sure why this is not working, but can describe what I've seen so far: 

the Initialize() function has to be put in either an ALWAYS or CLIENT System to show changes on both clients. That's the part that sets the HealthHud icon to White icons on start: 

```
    @Override
    public void initialise() {
        HealthHud healthHud = nuiManager.getHUD().getHUDElement("core:healthHud", HealthHud.class);
        healthHud.find("healthBar", UIIconBar.class).setIcon(Assets.getTextureRegion(LASUtils.getHealthIcon(LASUtils.WHITE_TEAM)).get());
        healthHud.setSkin(Assets.getSkin(LASUtils.getHealthSkin(LASUtils.WHITE_TEAM)).get());
    }
```

The next part is the part that changes the the HealthHud icons to either red hearts/black spades depending on team chosen: 
```
        String team = event.team;
        HealthHud healthHud = nuiManager.getHUD().getHUDElement("core:healthHud", HealthHud.class);
        healthHud.find("healthBar", UIIconBar.class).setIcon(Assets.getTextureRegion(LASUtils.getHealthIcon(team)).get());
        healthHud.setSkin(Assets.getSkin(LASUtils.getHealthSkin(team)).get());
```

If this is placed in an AUTHORITY system (ex. in TeleporterSystem.java) changes that are made to the HUD icons will only be reflected on the host (in a host + client system), regardless of which player is selecting the teleporter. If this is done in a CLIENT or ALWAYS system (ex. sending event to client system), the HUDs of any connected players will change regardless of which player is selecting the teleporter. 

It seems that this could be a bug in how NUIManager handles changing the skin of the HealthHud, but I'm not sure how exactly the NUIManager handles the UI skins, since this is different than how it handles other things such as health value, which is read from a component and thus updated correctly to only the appropriate player in multiplayer. I'm also not sure if there's a different way to target the HealthHud of a particular player rather than using NUIManager to change the skin of the HealthHud. 

To reiterate, the HUD swapping works on single player but does not work/sync properly on multiplayer. Intended functionality is when a player chooses a team, just their HealthHud updates to the appropriate symbols, with other players' HUDs staying however they were before.

Cervator:
Closing as completed by #79!

PR #78, opened 2019-03-30
Replicate Magic Fool to clients

darshan3:
NPC is now visible at both server and client side.

### How to test:
1. Host a server.
2. Interact with NPC.
3. Join the server using a 2nd client.
4. Interact with NPC.


darshan3:
Current Issue: When a client activates dialog, it shows up on both server as well as the client.

src/main/java/org/terasology/las/platform/GuidanceNpcProvider.java
```diff
@@ -36,19 +37,16 @@
...
             EntityStore entity = new EntityStore(chooseFactionNpc);
             entity.addComponent(new LocationComponent(new Vector3f(0, 61, 24)));
-            entity.addComponent(dialogComp);
             buffer.enqueue(entity);
 
             Prefab beaconMark = Assets.getPrefab("BeaconMark").get();
             EntityStore entityStore = new EntityStore(beaconMark);
             entityStore.addComponent(new LocationComponent(new Vector3f(0, 63, 24)));
-
+            entityStore.addComponent(new NetworkComponent());
```
jellysnake: Is there a reason for adding this component through code over prefab?

darshan3: 'BeaconMark' prefab is taken from Quests module. Hence, adding network component would require changing that module. Should I create beaconmark prefab for this module too?

jellysnake: I'd suggest looking into "delta"s. They allow you to modify a prefab from another module.
[See this guide for more info](https://github.com/Terasology/TutorialAssetSystem/wiki/Deltas-and-Overrides)

PR #82, opened 2019-05-31
Respawn and heal player on death

darshan3:
**Project Card**:
[Handle respawn on death](https://github.com/orgs/Terasology/projects/17#card-21937677)

**How to test**:  
4. Kill one of the players by repeatedly attacking it through another one.

**Expected outcome**:
The killed player should get respawned at its home base with full health restored.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
@@ -39,10 +39,10 @@ public class PlayerDeathSystem extends BaseComponentSystem {
     @ReceiveEvent(priority = EventPriority.PRIORITY_HIGH)
     public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterComponent characterComponent, AliveCharacterComponent aliveCharacterComponent) {
         if (player.hasComponent(PlayerCharacterComponent.class)) {
-            // Consume the BeforeDestroyEvent so that the DoDestroy event is never sent for player entities
             event.consume();
             logger.info(player.toFullDescription());
             String team = player.getComponent(LASTeamComponent.class).team;
+            player.send(new DoHealEvent(100000, player));
             player.send(new CharacterTeleportEvent(LASUtils.getTeleportDestination(team)));
         }
     }
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.ligthandshadow.componentsystem.controllers;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,49 @@
...
+public class PlayerDeathSystem extends BaseComponentSystem {
+    Logger logger = LoggerFactory.getLogger(PlayerDeathSystem.class);
+
+    @ReceiveEvent(priority = EventPriority.PRIORITY_HIGH)
+    public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterComponent characterComponent, AliveCharacterComponent aliveCharacterComponent) {
+        if (player.hasComponent(PlayerCharacterComponent.class)) {
+            event.consume();
+            logger.info(player.toFullDescription());
+            String team = player.getComponent(LASTeamComponent.class).team;
+            player.send(new DoHealEvent(100000, player));
```
llvieira: @darshan3 what does the magic number 100000 mean? it would be better to use a constant.

darshan3: I took it from healthMax command. https://github.com/MovingBlocks/Terasology/blob/develop/modules/Core/src/main/java/org/terasology/logic/health/HealthCommands.java

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,49 @@
...
+
+@RegisterSystem
+public class PlayerDeathSystem extends BaseComponentSystem {
+    Logger logger = LoggerFactory.getLogger(PlayerDeathSystem.class);
+
+    @ReceiveEvent(priority = EventPriority.PRIORITY_HIGH)
+    public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterComponent characterComponent, AliveCharacterComponent aliveCharacterComponent) {
+        if (player.hasComponent(PlayerCharacterComponent.class)) {
+            event.consume();
+            logger.info(player.toFullDescription());
```
nihal111: Remove this log before merge.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
nihal111: @darshan3 find out the value which restores the health completely and use that as a local constant for now.
Might ask @e-aakash to set that up as a constant somewhere in the health module and reference it later.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,49 @@
...
+            String team = player.getComponent(LASTeamComponent.class).team;
+            player.send(new DoHealEvent(100000, player));
+            player.send(new CharacterTeleportEvent(LASUtils.getTeleportDestination(team)));
```
nihal111: Ideally, we should not be using the exact location of the base. That would mean when multiple players die, they always land on top of each other. Since the base is big enough, we could teleport to a random position on the base as in `LASUtils.getTeleportDestination(team)) + Vec3(random(-1, 1), random(-1, 1), 0)`.

Again, here make sure that the players always face the opponent team's base. Maybe extract the above completely into a method and use it in the team selection logic as well.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
skaldarnar: The Health module should have a way to `restoreToMax` - not sure how/whether a constant can solve this.

nihal111: Sure, extracting the DoHeal event logic with the maxHealth value to such a method would be ideal.

e-aakash: If we make a constant in health component, we would anyway need to access the component to get the value. We can either use `maxHealth` for now or just this _hack_ of arbitrarily large value. 
`restoreFullEvent` is part of the present plan for the Restoration system. So, an event to restore to max health would be available when Restoration is complete (around the end of the second month of the coding period).

PR #83, opened 2019-06-01
Handle inventory on death

darshan3:
**Project Card**:
[Handle respawn on death](https://github.com/orgs/Terasology/projects/17#card-21937677)

**How to test**:  
4. Give "player one" items and blocks through cheat "give item"
5. Collect opposition flag using 'player one'.
4. Kill "player one" by repeatedly attacking it through "player two".
5. Try to collect dropped items and flag using "player two".

**Expected outcome**:
The killed player should get respawned at its home base with full health restored. All the items held by them should get dropped at the same location except the magic staff. Trying to collect the dropped flag sends it directly to home base.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
@@ -33,8 +33,13 @@ import org.terasology.logic.health.BeforeDestroyEvent;
 import org.terasology.logic.health.DoHealEvent;
 import org.terasology.logic.inventory.InventoryComponent;
 import org.terasology.logic.inventory.InventoryManager;
+import org.terasology.logic.location.LocationComponent;
 import org.terasology.logic.players.PlayerCharacterComponent;
+import org.terasology.math.geom.Vector3i;
 import org.terasology.registry.In;
+import org.terasology.world.WorldProvider;
+import org.terasology.world.block.BlockManager;
+import org.terasology.world.block.items.BlockItemComponent;
 
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,80 @@
...
+    @ReceiveEvent(priority = EventPriority.PRIORITY_HIGH)
+    public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterComponent characterComponent, AliveCharacterComponent aliveCharacterComponent) {
+        if (player.hasComponent(PlayerCharacterComponent.class)) {
+            event.consume();
+            String team = player.getComponent(LASTeamComponent.class).team;
+
+            Prefab staffPrefab = assetManager.getAsset(LASUtils.MAGIC_STAFF_URI, Prefab.class).orElse(null);
+
+            Vector3f startPosition = new Vector3f(player.getComponent(LocationComponent.class).getLocalPosition());
+            Vector3f impulse = Vector3f.zero();
```
iaronaraujo: Just out of curiosity, what would happen if the impulse wasn't 0?

darshan3: Impulse 0 drops the items at the same place. If we increasing its value, items will be thrown in that direction.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,80 @@
...
+            for (int slotNumber = 0; slotNumber <= inventorySize; slotNumber++) {
+                EntityRef slot = inventoryManager.getItemInSlot(player, slotNumber);
+                Prefab currentPrefab = slot.getParentPrefab();
+                if (currentPrefab != null && !currentPrefab.equals(staffPrefab)) {
+                    int count = inventoryManager.getStackSize(slot);
+                    player.send(new DropItemRequest(slot, player, impulse, startPosition, count));
+                }
+            }
+
+            player.send(new DoHealEvent(100000, player));
```
llvieira: Same here, it'd be better to use a constant variable.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,80 @@
...
+    @In
+    private WorldProvider worldProvider;
+
+    @In
+    private BlockManager blockManager;
+
+    @ReceiveEvent(priority = EventPriority.PRIORITY_HIGH)
+    public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterComponent characterComponent, AliveCharacterComponent aliveCharacterComponent) {
+        if (player.hasComponent(PlayerCharacterComponent.class)) {
+            event.consume();
```
nihal111: Add a comment above explaining that this event handler consumes the event before the player entity is destroyed, hence avoiding a death screen and respawning the player at base.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,80 @@
...
+            String team = player.getComponent(LASTeamComponent.class).team;
+
+            Prefab staffPrefab = assetManager.getAsset(LASUtils.MAGIC_STAFF_URI, Prefab.class).orElse(null);
+
+            Vector3f startPosition = new Vector3f(player.getComponent(LocationComponent.class).getLocalPosition());
```
nihal111: Why do we call this the `startPosition`? If anything, it should be the `endPosition` :laughing: or better, `deathPosition`.

skaldarnar: `deathPosition` sounds about right …

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
skaldarnar: which might be a nice effect later (to scatter them in a small radius around the player). Could even use the direction of killing impact … 🤔

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -0,0 +1,80 @@
...
+            Vector3f impulse = Vector3f.zero();
+            int inventorySize = inventoryManager.getNumSlots(player);
+            for (int slotNumber = 0; slotNumber <= inventorySize; slotNumber++) {
+                EntityRef slot = inventoryManager.getItemInSlot(player, slotNumber);
+                Prefab currentPrefab = slot.getParentPrefab();
+                if (currentPrefab != null && !currentPrefab.equals(staffPrefab)) {
+                    int count = inventoryManager.getStackSize(slot);
+                    player.send(new DropItemRequest(slot, player, impulse, startPosition, count));
+                }
+            }
```
skaldarnar: Could you please introduce a new method to keep the inventory handling in a concise spot. The `beforeDestroy` method should read as close to the following as possible (i.e., a high-level description of the semantics):

```java
public void beforeDestroy(…) {
    event.consume();
    dropItemsFromInventory(player, …);
    healPlayer(player);
    respawnPlayerAtBase(player);
}
```

`healPlayer` and `respawnPlayerAtBase` are already single lines of code, but for inventory a new method will help.

PR #89, opened 2019-06-11
Destroy dropped items on death after a delay

darshan3:
**Project Cards**:
1. [Handle Player Death [3]](https://github.com/orgs/Terasology/projects/17#card-22550695)

**How to test**:  
4. Give some blocks and prefab items to player 1.
5. Capture the flag with player 1.
6. Kill player 1 with player 2.
7. Collect some dropped items with player 2.
8. Repeat step 4 to 7 by exchanging player 1 and 2.

**Expected outcome**:
Items not collected will get destroyed. If it is flag, it will get teleported back to the base. If you collect the item, delay will be removed and it will not be destroyed.


jdrueckert:
I can confirm that the delayed item destruction works if the player is killed. However, it does not work if the flag is lost for some other reason, e.g. by getting hit with the activation key 'e'.
Would be great if not every event that leads to loosing the flag would need separate consideration for the delay, but if there'd be a general ItemDropped event or something alike that could be used to introduce the delay with.

darshan3:
Closing as an improved method has been implemented by #93!

PR #90, opened 2019-06-12
Health hud fix for reconnects and restarts

darshan3:
**Project Cards**:
1. [Handle loading from saved game and client reconnects](https://github.com/orgs/Terasology/projects/17#card-22540317)

**How to test**:  
4. Leave the game for one of the clients and join in again.
5. Close the game and try loading the game back from save and join using other clients again.

**Expected outcome**:
Health HUD should be initially white dots and it should change to Spades for black team, hearts for red team.


skaldarnar:
This does work for a single client re-joining a game, and for the host when quitting the game and then hosting it again (loading the same game). 
However, it does not work for a previously connected client if they lost connection because the host quit. Player appears as _white_ and also has the white dots.

Terasology-190614000627-1152x720
image

skaldarnar:
On the upside, it works if the client disconnects before the host. When re-joining after the host is up again the HUD is correct. 
What's the difference between a graceful disconnect and losing connection to the server?

PR #104, opened 2019-08-15
Integrate minimap with LASTeam based icons

darshan3:


iaronaraujo (approved):

src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
```diff
@@ -51,6 +51,8 @@
     public static final String WHITE_HEALTH_SKIN = "lightAndShadowResources:healthWhite";
     public static final String RED_HEALTH_SKIN = "lightAndShadowResources:healthRed";
     public static final String BLACK_HEALTH_SKIN = "lightAndShadowResources:healthBlack";
+    public static final String RED_MINIMAP_ICON = "lightAndShadowResources:heartsIcon";
```
skaldarnar, line 54: As pointed out in https://github.com/Terasology/LightAndShadowResources/pull/31#pullrequestreview-277364660 you might just use the icons that are already there:
```suggestion
    public static final String RED_MINIMAP_ICON = "lightAndShadowResources:heart";
```

src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
```diff
@@ -51,6 +51,8 @@
...
+    public static final String BLACK_MINIMAP_ICON = "lightAndShadowResources:spadesIcon";
```
skaldarnar, line 55: As pointed out in https://github.com/Terasology/LightAndShadowResources/pull/31#pullrequestreview-277364660 you might just use the icons that are already there:
```suggestion
    public static final String BLACK_MINIMAP_ICON = "lightAndShadowResources:spade";
```

skaldarnar (changes requested):

jdrueckert (approved):

PR #106, opened 2019-08-20
Revert "fix(LightAndShadow): replace perlin with simplex"

jdrueckert:
Reverts Terasology/LightAndShadow#102
Perlin and Simplex seem to differ in their noise functions. Until we know how to correctly configure Simplex to act like Perlin, this change is reverted and the removed Perlin parts reintroduced and marked as deprecated.


skaldarnar:
Accompanying https://github.com/MovingBlocks/Terasology/pull/3730

PR #118, opened 2019-10-19
Update structural resources dependency to 2.0.0

jdrueckert:


skaldarnar:
If we are not using any of the new blocks we don't have to update this dependency, do we?

PR #125, opened 2020-06-29
Ammo Override UI

stefaniamak:
Associated PR: [Test Assets for LightAndShadow](https://github.com/Terasology/LightAndShadowResources/pull/40)

# Result

The quiver HUD gets Overriden.

## Changes

* Box moved a little bit to the right
* "Ammo" label placed above the box

### Issues

- [X] ~~Ammo box doesn't appear when the game begins (if an item with ammo is the first selected), but you have to change the item and then select back the one with ammo for it to appear~~ **-- Worked on https://github.com/Terasology/CombatSystem/pull/50** 
- [X] ~~I need to figure out how to manipulate label type elements better~~ **-- Resolved within the comment section https://github.com/Terasology/LightAndShadow/pull/125#issuecomment-657079106** 

## How to test

1. Make sure you use the test assets from this `draft` PR: https://github.com/Terasology/LightAndShadowResources/pull/40 
2. Run a LaS game
3. Select the `bow` at quickslot

## Test UI screenshot

Screenshot_4

### DIFF
--- a/assets/skins/ammoHud.skin
+++ b/assets/skins/ammoHud.skin
@@ -2,7 +2,7 @@
     "inherit": "default",
     "elements": {
         "InventoryCell": {
-            "background": "LightAndShadowResources:ammoArea",
+            "background": "engine:area",
             "background-border": {
                 "top": 4,
                 "bottom": 4,
@@ -17,7 +17,7 @@
             },
             "modes": {
                 "active": {
-                    "background": "LightAndShadowResources:boxActive"
+                    "background": "engine:boxActive"
                 }
             }
         },
--- a/assets/skins/ammoHud.skin
+++ b/assets/skins/ammoHud.skin
@@ -1,0 +1,41 @@
+{
+    "inherit": "default",
+    "elements": {
+        "InventoryCell": {
+            "background": "LightAndShadowResources:ammoArea",
+            "background-border": {
+                "top": 4,
+                "bottom": 4,
+                "left": 4,
+                "right": 4
+            },
+            "margin": {
+                "top": 4,
+                "bottom": 4,
+                "left": 4,
+                "right": 4
+            },
+            "modes": {
+                "active": {
+                    "background": "LightAndShadowResources:boxActive"
+                }
+            }
+        },
+        "TransferItemCursor": {
+            "align-horizontal": "center",
+            "align-vertical": "top"
+        },
+        "ItemIcon": {
+            "fixed-width": 48,
+            "fixed-height": 48,
+            "texture-scale-mode": "scale fit",
+            "text-align-horizontal": "right",
+            "text-align-vertical": "bottom"
+        },
+        "UICrosshair": {
+            "background": "engine:gui#crosshair",
+            "fixed-width": 40,
+            "fixed-height": 40
+        }
+    }
+}
--- a/overrides/CombatSystem/ui/hud/quiverHud.ui
+++ b/overrides/CombatSystem/ui/hud/quiverHud.ui
@@ -1,6 +1,6 @@
 {
     "type": "QuiverHud",
-    "skin": "inventoryDefault",
+    "skin": "ammoHud",
     "contents": {
         "type": "relativeLayout",
         "contents": [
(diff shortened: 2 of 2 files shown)


jdrueckert:
> Ammo box doesn't appear when the game begins

Ideally, we should fix this. Knowing our code base and how difficult it can be to track something like this down, I'll propose that you take a time slot of 30-60 minutes max(!) to try to figure it out. If you cannot fix it in that time slot, please document the issue and your findings in a bug ticket ("Issue") on LaS and work around this issue by changing the default item selected on game start (that should be easier).

> I need to figure out how to manipulate label type elements better

Can you elaborate a bit on that? What about the current manipulation (actions or possibilities?) of label type elements is not yet good enough?

stefaniamak:
> > I need to figure out how to manipulate label type elements better
> 
> Can you elaborate a bit on that? What about the current manipulation (actions or possibilities?) of label type elements is not yet good enough?

@jdrueckert 

I currently can't really figure out how to change the size of the font, and generally, if it's possible to change other "text" features like boldness, align, etc..

jdrueckert:
Ahh.. so basically we have to find out how to do text formatting in `UILabel`...?

stefaniamak:
Yes, exactly! The elements seem to be documented [here](https://github.com/Terasology/TutorialNui/wiki/Core-Widgets), but there is no other attribute other than the `text` under `UILabel`.

jdrueckert:
Okay, wild guess: did you try HTML tags? 🤔 
I think they work in other text properties, e.g. if you take a look at the book in MOO for instance.
IIRC we have a custom HTML style there, though, so for example `<l>` instead of `<br>` for a line break - might be that the usual text formatting tags like `<i>` or `<b>` don't work, but it's worth a try I'd say

stefaniamak:
> Okay, wild guess: did you try HTML tags? 🤔

I see 🤔 didn't think of that at all. I am looking into it atm, and it seems a bit tricky generally. I'll proceed with other tasks before looking more into this matter probably.

I did find that within the `MagicFool.prefab` (which include The Fool's text) there is the option to edit the text - it does change the size of `"Welcome"` text by setting it as `title`. So maybe text should be done through prefabs? Then we would have to change the design's code twice though, like this (The Fool's text) case. 
I tried to put `"title"` or `"h1"` in the same style as it is within the `prefab`, but didn't work.

jdrueckert:
> I tried to put "title" or "h1" in the same style as it is within the prefab, but didn't work.

Okay, so you could try to find out which nui widget is used for the MOO book and compare the implementation of that nui widget with the implementation of `UILabel` - might be that the label is currently implemented to be plain text only

stefaniamak:
Based on my search, there is a module named `Book`, though there is non `UILabel` used into any of its UI files. Also in MOO the only book-related code as I can see is done through `prefab` files.

jdrueckert:
What kind of formatting did you have in mind for the `UILabel` text and for what reason/purpose? Maybe we can find another way to achieve a similar effect without having to introduce formatting options to `UILabel` if there's none present yet 🤔

stefaniamak:
The main thing I needed was the size of the text, and the purpose was to use it for titles. The rest like **bold** or _italic_ I believe is secondary to have.

Cervator:
Tiny site-note about stylized text: with internationalization my understanding is that italics and underlining don't make sense in some language scripts anyway, leaving just bold potentially meaningful. And maybe colors.

stefaniamak:
> Tiny site-note about stylized text: with internationalization my understanding is that italics and underlining don't make sense in some language scripts anyway, leaving just bold potentially meaningful. And maybe colors.

Indeed actually. And now that I think about it, we should use the skin files for these modifications as well; e.g. how we changed the size of the health bar through the different skins (https://github.com/Terasology/LightAndShadowResources/pull/42).

With that conclusion, I believe the issue is resolved. Thanks for your comments!

jdrueckert:
https://github.com/Terasology/CombatSystem/pull/50 merged

jdrueckert:
> With that conclusion, I believe the issue is resolved.

Alright then, are there more overrides you want to do for the ammo slot and if so, which would that be? Ideally, please update the PR description with a checkbox list using
```
- [ ] <item>
```
which should result in the following style
- [ ] item

so we can track what's done and what's still missing.

If there's nothing missing, then please make this a real (non-draft) PR and remove the references to the test assets.

jdrueckert:
@stefaniamak I'm afraid, this cannot yet be merged.
image
As you can see the active and ammo slots don't have an image anymore...

jdrueckert (approved):

Issue #130, opened 2020-07-20
Game OverScreen Overhaul


jdrueckert:
## Desired result
![image](https://user-images.githubusercontent.com/48293545/87904267-98f6af80-ca66-11ea-8002-c03d5f7f69e0.png)

## Tasks
- [x] Move skin file from LaS repo to LaSR repo - done in #128 
- [X] Update reference to skin file - done in #128 
- [x] Place all elements inside the shown boxes at the screenshot above - done in #136
- [x] Set `Game Over` and `Victory`/`Defeat` (currently `You Win!` and `You Lost!`(?)) UILabels to the top left - done in #136
- [x] Add `Hearts Team` and `Spades Team` labels - done in #136
- [x] Format spacing between details in the middle of the screen - done in #136
- [x] Add final score result _(the `Enemy Flags Captured` and points below that label)_ - done in #136
- [x] Change position of `Restart`, `Exit to Main Menu`, `Exit Terasology` and `Settings` buttons, based on the screenshot below  - done in #136
- [x] Add team symbol images before `Hearts Team` and `Spades Team` - done in #146 & https://github.com/Terasology/LightAndShadowResources/pull/56
- [x] Add wider images close to the bottom of the wider of the two boxes - done in #136
- [x] Set font sizes accordingly - done in #146 & https://github.com/Terasology/LightAndShadowResources/pull/56
- [x] Set new background images - done in #146 & https://github.com/Terasology/LightAndShadowResources/pull/56
- [x] Change button's design, according to the background, font color ect.. (?) - done in #146 & https://github.com/Terasology/LightAndShadowResources/pull/56
- [x] Add different font sizes - done in https://github.com/Terasology/LightAndShadowResources/pull/56
- [x] Asset, buttons and element placement - done in https://github.com/Terasology/LightAndShadow/pull/146 & https://github.com/Terasology/LightAndShadowResources/pull/56
- [x] Change font-color for `Victory`-`Defeat` label - done in https://github.com/Terasology/LightAndShadowResources/pull/57 & https://github.com/Terasology/LightAndShadow/pull/148
- [x] Add a border to outside UIBox - done in https://github.com/Terasology/LightAndShadow/pull/149 & https://github.com/Terasology/LightAndShadowResources/pull/58
- [ ] Place player names at the start of the column, and not at the middle of it
- [ ] Add point assets
- [ ] Add card assets that indicate the `Enemy Flags Captured`, like the mock-up

## How to test
1. Start a LaS game
[...]

PR #136, opened 2020-07-23
feat: game over screen rework

stefaniamak:
## Purpose

This PR will hold commits regarding tasks from issue Game Over Screen Overhaul https://github.com/Terasology/LightAndShadow/issues/130.

## How to test

1. start a LAS game
2. score 5 points by taking the enemy flag, and bringing it to your base*
3. the screen shall appear when the last point is scored

*you can score points by clicking the box  on the enemy base. To avoid scoring 5 points, you can go to file `LASUtils.java`, and change `GOAL_SCORE = 5` to `GOAL_SCORE = 1`, then you would only need to score 1 point.

### By checking and the above `How to test` steps, you will get the following result
image

### MigLayout areas Highlighted

Not necessary for checking, but if you want to get the MigLayout areas highlighted, you can replace the `gameoverScreen.skin`'s code with the following lines:
```
{
    "inherit": "default",
    "elements": {
        "UIScreenLayer": {
            "background-scale-mode": "stretch",
            "background": "engine:greyedBackground"
        },
        "UIButton": {
            "min-width": 128,
            "min-height": 32
        },
        "UILabel": {
            "font" : "lightAndShadowResources:WickedQueen"
        },
        "MigLayout": {
            "background-scale-mode": "stretch",
            "background": "engine:greyedBackground"
        }
    }
}
```

image


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
```diff
@@ -86,19 +87,27 @@ public void onGameOver(GameOverEvent event, EntityRef entity) {
...
             for (EntityRef character : characters) {
                 EntityRef client = character.getOwner();
                 ClientComponent clientComponent = client.getComponent(ClientComponent.class);
-                migLayout.addWidget(new UILabel(PlayerUtil.getColoredPlayerName(clientComponent.clientInfo)), new MigLayout.CCHint());
+                String playerTeam = localPlayer.getCharacterEntity().getComponent(LASTeamComponent.class).team;
                 PlayerStatisticsComponent playerStatisticsComponent = character.getComponent(PlayerStatisticsComponent.class);
-                migLayout.addWidget(new UILabel(String.valueOf(playerStatisticsComponent.kills)), new MigLayout.CCHint());
-                migLayout.addWidget(new UILabel(String.valueOf(playerStatisticsComponent.deaths)), new MigLayout.CCHint("wrap"));
+                migLayout = (playerTeam.equals("black") ? spadesTeamMigLayout : heartsTeamMigLayout);
+                addInfoToTeamMigLayout(migLayout,clientComponent,playerStatisticsComponent);
```
stefaniamak: I was in between two ways of implementing this:

1. have an extra migLayout variable, and hold there the migLayout that the playerDetails are gonna be added at: _(this is also what I am doing at the marked lines)_
```JAVA
migLayout = (playerTeam.equals("black") ? spadesTeamMigLayout : heartsTeamMigLayout);
addInfoToTeamMigLayout(migLayout,clientComponent,playerStatisticsComponent);
```
2. have an if statement and repeat the call of `addInfoToTeamMigLayout` twice, but with a different migLayout sent each time:
```JAVA
if (playerTeam.equals("black")){
     addInfoToTeamMigLayout(spadesTeamMigLayout, clientComponent, playerStatisticsComponent);
} else {
     addInfoToTeamMigLayout(heartsTeamMigLayout, clientComponent, playerStatisticsComponent);
}
```

With _1._ we avoid repeating the same code (the call of addInfoToTeamMigLayout) and the code looks more clear and simple, while with the _2._ we avoid having an extra variable. I believe the _1._ is a better option, thus why I went with it, but I want to hear your opinion as well @skaldarnar and @jdrueckert .

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
```diff
@@ -86,20 +87,38 @@ public void onGameOver(GameOverEvent event, EntityRef entity) {
...
+
+    private void addTeamScore(DeathScreen deathScreen){
+        UILabel spadesTeamScore = deathScreen.find("spadesTeamScore", UILabel.class);
+        UILabel heartsTeamScore = deathScreen.find("heartsTeamScore", UILabel.class);
+        ClientScoreSystem clientScoreSystem = new ClientScoreSystem();
+        String blackScore = clientScoreSystem.getBlackScore();
+        String redScore = clientScoreSystem.getRedScore();
+        spadesTeamScore.setText(blackScore);
+        heartsTeamScore.setText(redScore);
+    }
```
stefaniamak: Here I get the score number of `0` for both teams. I believe this is probably because the score elements get reset somewhere in between when the game ends and when the scores get showcased.
image

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientScoreSystem.java
```diff
@@ -63,6 +63,14 @@ public String get() {
         redGoalScore.setText(Integer.toString(LASUtils.GOAL_SCORE));
     }
 
+    public String getBlackScore(){
+        return String.valueOf(blackScore);
+    }
+
+    public String getRedScore(){
+        return String.valueOf(redScore);
+    }
```
stefaniamak: Not sure if adding these getters is the best way to solve the issue.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
```diff
@@ -86,20 +87,38 @@ public void onGameOver(GameOverEvent event, EntityRef entity) {
...
-        if (migLayout != null) {
+        MigLayout spadesTeamMigLayout = deathScreen.find("spadesTeamPlayerStatistics", MigLayout.class);
+        MigLayout heartsTeamMigLayout = deathScreen.find("heartsTeamPlayerStatistics", MigLayout.class);
+        if (spadesTeamMigLayout != null && heartsTeamMigLayout != null) {
             Iterable<EntityRef> characters = entityManager.getEntitiesWith(PlayerCharacterComponent.class, LASTeamComponent.class);
             for (EntityRef character : characters) {
                 EntityRef client = character.getOwner();
                 ClientComponent clientComponent = client.getComponent(ClientComponent.class);
-                migLayout.addWidget(new UILabel(PlayerUtil.getColoredPlayerName(clientComponent.clientInfo)), new MigLayout.CCHint());
+                String playerTeam = localPlayer.getCharacterEntity().getComponent(LASTeamComponent.class).team;
```
stefaniamak: Not sure how `localPlayer.getCharacterEntity()` works on multiplayer.

Cervator: Have you experimented with it in the different cases? The area can indeed be a bit tricky, between single player, headless servers, hosting clients, and purely joining clients

With `ClientGameOverSystem` being a `CLIENT` system you should be fine vs headless servers. The `localPlayer` should be _something_ in all three other cases, but it is hard to remember any subtle differences

stefaniamak: > The localPlayer should be something in all three other cases

Do you mean it might be something different in each loop?

skaldarnar:
No review from me this time, just wanted to give a 👍 for making progress visible by pushing all these small commits 🤓 

... off to vacation 🥾 🌲 ⛺

overrides/engine/ui/ingame/deathScreen.ui
```diff
@@ -1,112 +1,328 @@
...
+              "type": "RowLayout",
+              "horizontalSpacing": 15,
+              "contents": [
+                {
+                  "type": "UILabel",
+                  "text": "Game Over"
+                },
+                {
+                  "type": "UILabel",
+                  // "You Win!" "You Lose!" label
```
Cervator: Normally I love comments, but JSON isn't a big fan of them - our parser is set to lenient mode where it'll ignore them, but some tools might whine and cry about comments in JSON. And this one can probably be expressed well enough in the `id` to where we could lose the label but still win :-)

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
```diff
@@ -56,16 +56,17 @@
     /**
      * System to show game over screen once a team achieves goal score.
      *
-     * @param event  the event
+     * @param event the event
```
Cervator: I realize you just caught this line by fixing an errant space, but now that you've touched it anyway want to take a quick stab at adding a few more words for each param to make them more meaningful? 😁

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
Cervator: Sorry, just meant that in each way you can run the game the setup would differ. Per client. Not anything related to a loop. There should only ever be one `localPlayer` but potentially multiple _actual_ players :-)

I was able to test around a bit, I crashed headless but suspect that was from running multiple clients from one workspace in funny ways 😅

Cervator (dismissed):
This works out and the code looks pretty solid! Nicely done.

I left a couple minor comments and have one open question, none really prevent merging this so we can either do so first time after this we're able to, or you could throw one additional tiny commit at it if desired first :-)

The open question goes with the screenie below: Is it intentional that player names are in red and _green_ ? I could have easily missed some discussion on that earlier. That threw me off, although at first I only had two players so I thought maybe they were just randomized or player-chosen colors. Then I ran with 6 to test scrolling and it was all green with one red. Well, two reds, as I think I overlapped some stuff in my workspaces (somehow got 10 player entries after one score reset), that's probably not related to this PR 😅 

My expectation would probably either be black names for the black team, red names for red. Maybe if we want to mark the active player they could be in white? Or green for that sake.

Another thought: Would it make sense to put a white outline around the winning team's logo in the bottom left or right? Not at all a necessity for this PR, just an idea

2020-08-14 00_37_03-Terasology _ Alpha

overrides/engine/ui/ingame/deathScreen.ui
stefaniamak: Oh! Yes yes I see, I'll remove it!

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
stefaniamak: Yes, will do so!

stefaniamak:
**About the player name colors:** the names are supposed to be the color that the player sets for them at the settings! I did not edit this feature myself, but I both stumbled upon that code while working on the screen and had the colors I had placed to my clients' display to their names. _(the image attached shows the colors in my build while running 3 clients. It's an earlier version of my testing, thus the reason why the structure of the screen is not the latest one)_

My initial thought about the team name colors was to have them all in one color, except the user's username. I do also like your idea about having them red for the hearts team and black for the spades team, it really does depend on the overall color placement and scheme of the whole game over screen - so if we end up using these text colors, we would have to build the rest of the screen based on that. Considering we don't have any specific coloring for the screen yet, all ideas are welcomed!

**About the white outline:** I think that depends on the final background 🤔 But now that you mention it, the hearts icon seems kind of odd without an outline, in comparison with the spades icon which has a black border.

**I have an extra question about your screenshot:** did the MigLayout with the names for the Spades Team grow while adding more player names? (I say "adding more" because there is an error where if you click restart and complete the game again, then the names would be repeated) . I ask because it is supposed to keep a specific height, and in my build, the size does not change no matter the names.

image

Cervator (approved):
Did one last test after looking at the final commits - merging! :-)

One other observation: For me at least with default config the right-side tool bar slots actually overflow off the top of the screen, with only like 8.5 slots visible. Is that just me or a known issue? The slots look a tad large.

Also, a few log entries print after the game over screen pop up, any idea why?

`WARN  o.t.rendering.nui.asset.UIFormat - Field 'cc' not recognized for interface org.terasology.rendering.nui.UIWidget in {"type":"UILabel","text":"Kills","cc":""}`

stefaniamak:
> 
> 
> Did one last test after looking at the final commits - merging! :-)
> 
> One other observation: For me at least with default config the right-side tool bar slots actually overflow off the top of the screen, with only like 8.5 slots visible. Is that just me or a known issue? The slots look a tad large.
> 
> Also, a few log entries print after the game over screen pop up, any idea why?
> 
> `WARN o.t.rendering.nui.asset.UIFormat - Field 'cc' not recognized for interface org.terasology.rendering.nui.UIWidget in {"type":"UILabel","text":"Kills","cc":""}`

@Cervator 

**Quickslot being covered**

The quickslot items were made slightly bigger in a previous PR, which's purpose was to avoid overlapping of the slot's borders because of the new, wider quickslot's background asset. We would either widen the slot sizes (what we ended up doing) or widen the horizontal spacing to avoid this collision. Niruandaleth had mentioned some while ago that the space there is between the item inside the selected slot and its border is really nice, so we decided to go with the option of enlarging all the slots just slightly, to both solve our issue and provide some of that space within the slot to the rest of the slots too.

I suspected that there wouldn't be someone who played terasology in such a window that the total height of the quickslot won't be enough 🙈 I do believe that we will have this issue nonetheless, because there is always going to be a too-small of a screen that covers part of the UI.

**log entries**

I was not aware of that! I'll make sure to check it out!

Issue #131, opened 2020-07-20
Score HUD Element Overhaul


jdrueckert:
## Desired result
![20200723_105544](https://user-images.githubusercontent.com/48293545/88284767-9f9b5600-ccf6-11ea-8a58-5285bc6af172.jpg)

## Tasks
- [x] Move skin file from LaS repo to LaSR repo - done in #129 
- [X] Update reference to skin file - done in #129 
- [x] Create a new design 
- [x] Place UIImage at the middle - WIP #138 
- [x] Place UILabels next to the current team scores, and connect it with `GOAL_SCORE` at `LASUtils.java` - WIP #138 
- [x] Place UI images next to the scores, indicating which score is which team's - WIP #138 
- [X] Create a PR for LASR, with the task to create and insert the Flag image in the middle of the UI - done https://github.com/Terasology/LightAndShadowResources/pull/47

## How to test
1. Start a LaS game
2. Notice the points at the top middle of the screen

skaldarnar:
I think this one can be closed as **completed**!

PR #138, opened 2020-07-24
feat: Score hud rework

stefaniamak:
## Tasks left

- [x] Place the element correctly in the middle of the screen
- [x] Add more spacing between the team name and the score
- [x] text shadows for the team names and scores - WIP https://github.com/Terasology/LightAndShadowResources/pull/52
- [x] use the heart icon color as the font color for the "Hearts" team name - WIP https://github.com/Terasology/LightAndShadowResources/pull/52

## Purpose

This PR will hold commits regarding tasks from issue **Score HUD Element Overhaul https://github.com/Terasology/LightAndShadow/issues/131**.

## Additions included so far

- Added a UIImage in between the scores of the two teams _(ToDo: set new asset when that is committed at LASR)_
- Added team images left and right of the element 

image

- Added UILabels next to the current team scores, and 
- Connected the above UILabels with `GOAL_SCORE` at `LASUtils.java`

image

- Added new asset, and set layout according to it - asset at https://github.com/Terasology/LightAndShadowResources/pull/47
- Removed "Score" text from the UILabels
- Move "Red" UILabel right before the heart symbol at the right end of the element

image

- Set equal `width`s for `Black` and `Red` UILabels
- Changed UILabels `Black` to `Spades` and `Red` to `Hearts
- Added more spacing between the team name and the score
- Placed the element correctly in the middle of the screen

image

## How to test

1. get LASR latest version
2. start a LAS game
3. notice the Score Element at the middle top part of the screen

image


assets/ui/scoreHud.ui
```diff
@@ -7,7 +7,7 @@
             {
                 "type": "UIBox",
                 "layoutInfo": {
-                    "width": 300,
+                    "width": 400,
                     "height": 100,
                     "position-horizontal-center": {},
```
stefaniamak: When replacing the lines about `width` and `height` with
```JSON
"use-content-width": true,
"use-content-height": true,
```
the whole HUD element moves to the left, despite the `"position-horizontal-center": {},`, which is kind of confusing 🤔

stefaniamak: My first guess is that this probably has to do with the outside `relativeLayout` not having `layoutInfo`, but I haven't looked too much into it yet.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientScoreSystem.java
```diff
@@ -57,6 +57,10 @@ public String get() {
                 return String.valueOf(redScore);
             }
         });
+        UILabel blackGoalScore = scoreScreen.find("blackGoalScore", UILabel.class);
+        blackGoalScore.setText(Integer.toString(LASUtils.GOAL_SCORE));
+        UILabel redGoalScore = scoreScreen.find("redGoalScore", UILabel.class);
+        redGoalScore.setText(Integer.toString(LASUtils.GOAL_SCORE));
```
stefaniamak, line 63: I am not sure if this is the best solution to go about. It does work, and currently, we do not have an option to change the max amount of points needed in-game, so binding the `UILabel`s with the `GOAL_SCORE` variable is not _that_ necessary atm in my perspective. I would appreciate any opinions on this.

stefaniamak: Basically the above statement in more detail:
Here I read the two variables (one for Red Team's score and one for Black Team's score) the UILabel that holds the max amount of points you need to score for the game to end, which is set at variable `GOAL_SCORE` within `LASUtils.java`.

What I am doing above, is taking the variable's value directly from `GOAL_SCORE` and giving it to the UILabels, once _(I believe)_ when the Score HUD element is created.

My above comment was about the thought that the above method was not ideal, considering (e.g.) the need for that variable to change sometime in-game - but, I do not think there is any feature that changes that variable in-game, thus no need to use the binding method that the score uses _(see `onScoreUpdateFromServer` in `ClientScoreSystem.java`)_.

jdrueckert (changes requested):
That's what it currently looks like in-game:
image


Let's drop the "Score:" text on both sides and have the following structure:
:spades: "Black" `score-black` `flag icon` `score-red` "Red" :heart:
-> switch places of "Red" text and red score
The central icon can be replaced by a simple flag icon in this first iteration (by adding the asset in a separate PR in LASR and referencing it here) - doesn't need to be particularly pretty yet.

-- edited due to unclean workspace on my side that produced issues at first.

stefaniamak:
> 
> 
> That's what it currently looks like in-game:
> image
> 
> Let's drop the "Score:" text on both sides and have the following structure:
> ♠️ "Black" `score-black` `flag icon` `score-red` "Red" ❤️
> -> switch places of "Red" text and red score
> The central icon can be replaced by a simple flag icon in this first iteration (by adding the asset in a separate PR in LASR and referencing it here) - doesn't need to be particularly pretty yet.
> 
> -- edited due to unclean workspace on my side that produced issues at first.

Thanks for the suggestions! I was planning on removing at least the "Score" text, but positioning the "Red" and "Black" labels seem pretty nice! I'll do that :)

edit: actually, should we also change the team names from "Black" and "Red" to "Spades" and "Hearts" maybe?

stefaniamak:
So I meant for the flags' colors to be on the same side when creating the asset, but actually that opposite placement can both mean that "each team holds their flag" (the sticks of each flag is on their team's side) and provide a small "chaos" (by alternating the color order) which kinda fits theoretically and aesthetically to this Alice in Wonderland theme 🤔

I believe I prefer having the red flag to the left and black to the right, but what do you think?

**Both versions:**

**A)** opposite, red left black right
image

**B)** matching, black left red right
image

jdrueckert:
- flag icon (nice btw!): version B please as yes it fits the theme, but we should try to avoid confusion/chaos (even if it fits the theme)
- Black/Red vs. Spades/Hearts: don't care, do as you like, but please don't rack your brains about it - if you cannot decide on either, just leave as is ;)
- additional feedback: IMO we need a bit more spacing between the team name and the score - seems a bit plastered at the moment.

jdrueckert (changes requested):
As discussed, the remaining tasks for this PR are:
- text shadows for the team names and scores
- using the heart icon color as the font color for the "Hearts" team name

assets/ui/scoreHud.ui
```diff
@@ -7,22 +7,39 @@
...
+                                    "offset": 12
+                                }
+                            }
+                        },
                         {
                             "type": "UILabel",
                             "id": "blackScoreLabel",
                             "family": "blackText",
-                            "text": "Black Score: ",
+                            "text": "Spades  ",
```
jdrueckert: I believe we shouldn't work with whitespaces here but rather with an offset.

assets/ui/scoreHud.ui
```diff
@@ -61,10 +111,76 @@
...
+                                    "widget": "redSlash",
+                                    "target": "RIGHT"
+                                }
+                            }
+                        },
+                        {
+                            "type": "UILabel",
+                            "id": "redScoreLabel",
+                            "family": "redText",
+                            "text": "  Hearts",
```
jdrueckert: > I believe we shouldn't work with whitespaces here but rather with an offset.

Same here.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientScoreSystem.java
jdrueckert: One option you could go for here is to implement a kind of "getter" for the goal score, that currently returns the value of the `GOAL_SCORE` constant and can - once this is implemented - return the configured goal score instead. Pinging @skaldarnar for second opinion on this option.

jdrueckert (dismissed):

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientScoreSystem.java
skaldarnar: I think using the constant is a viable first approach here. I just double checked, there is an old `ScoreComponent` which seems to be unused right now. That component could be reused and extended with the target score for winning a game and other configuration. 

That falls into the bucket of cleaning up the package structure of light and shadow and doing a general refactoring/house keeping 🙈

skaldarnar (dismissed):

skaldarnar (approved):

PR #146, opened 2020-08-20
feat: game over screen UI rework

stefaniamak:
## Purpose

This PR is about modifications according to the design of the Game Over Screen.

## Changes made

* Added a `family` part at the elements within the UI file at the 

## How to test

1. Check https://github.com/Terasology/LightAndShadowResources/pull/56
2. Start a LAS game
3. Choose a team
4. Open the developer tab and gain the enemy team's by writting `give` and `redflag` or `blackflag`
5. Score as many points as you need to finish up the game
6. The Game Over Screen will pop up when you score the last point

_Note:_ You can go to `LAUtils.java` and change `GOAL_SCORE` variable from `5` to `1`, so you need to score 1 point to end the game.

## Current Version

image

### DIFF
--- a/overrides/engine/ui/ingame/deathScreen.ui
+++ b/overrides/engine/ui/ingame/deathScreen.ui
@@ -23,7 +23,7 @@
               "contents": [
                 {
                   "type": "UILabel",
-                  "text": "Game Over"
+                  "text": "Game Over:"
                 },
                 {
                   "type": "UILabel",
@@ -129,42 +129,34 @@
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
                         "type": "UISpace",
-                        "family": "noBackground",
                         "cc": ""
                       },
                       {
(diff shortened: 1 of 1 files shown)


overrides/engine/ui/ingame/deathScreen.ui
```diff
@@ -106,27 +117,67 @@
...
+                        "cc": ""
+                      },
+                      {
+                        "type": "UISpace",
+                        "cc": ""
+                      },
+                      {
+                        "type": "UISpace",
+                        "cc": ""
+                      },
```
stefaniamak, line 162: The only thing I didn't manage to reduce is this code. I tried many things, but non prove effective.

What I am doing here, is filling up the row with empty spaces, so the width of the row is forced to be divided into many little spaces. 

My initial tries revolved around the setting of 1 UISpace, and tried to set the desired widths via code, my tries - which all ended up at having 3 equally divided columns - were:
- Set a "width" at the UISpace
- Set ColumnConstrains at the MigLayout
- Place them in a RowLayout, and tried to modify that
- Use "span 8" at the UISpace to span it 8 cells
- Use "cell [column] [row] [width] [height]" where the "[ ]" contain numbers, the cells were arranged correctly, but the width and height didn't change
- Removed the "dock south" from the second MigLayout (L167) and placed it at the bottom with a width 3 and re-done the above methods, all failed

These are the methods that I can recall trying, and this above one was _sadly_ the only one that worked.

Cervator: Huh, interesting. Maybe be worth filing a bug report or feature enhancements on this into the TeraNUI repo?

Cervator (approved):
Tests out and codey bits look fine :-)

At this point it seems like there are only a few trivialities left on the screen, which can be handled in the future. Like some odd logging, the slight alignment of the flag capture stats (not centered with player box) or the buttons on the edge (like "Settings" or "Exit Terasology" appearing like they're partially _off_ the edge (too little margin?). Not really worried about those. Screen looks great! :-)

image

overrides/engine/ui/ingame/deathScreen.ui
stefaniamak: I could do that, yes. I just had a thought that this might occur only with UISpaces, considering the above methods mentioned were things that worked under others circumstances (uncertain what the others consist of).

stefaniamak:
> 
> 
> Tests out and codey bits look fine :-)
> 
> At this point it seems like there are only a few trivialities left on the screen, which can be handled in the future. Like some odd logging, the slight alignment of the flag capture stats (not centered with player box) or the buttons on the edge (like "Settings" or "Exit Terasology" appearing like they're partially _off_ the edge (too little margin?). Not really worried about those. Screen looks great! :-)
> 
> image

I didn't have the issue with the buttons appearing off the edge 😮 But yes, I thought it would be better to work the further changes in a new PR!

PR #147, opened 2020-08-24
feat migrate nui version/1.x

pollend:
issue: https://github.com/MovingBlocks/Terasology/issues/2307


Cervator:
Not sure what happened to this PR, but it seemingly just moved a single file, rather than do any NUI fixes 🤔 

Applying the fixes was easy enough, submitting via separate PR and closing this one. I'm leaving the one existing package issue alone (`UI` instead of `ui` - easy fix but not NUI so maybe fix later)

PR #158, opened 2020-09-01
Update Death Counter Due To Fall Damage

### LINKED ISSUE #145: Unable for a player to die if not killed by another player
## Issue

When the player reaches 0 HP alone without being killed by another client (e.g. when they fall from the high ground), they do not die but continue playing normally.

## Proposed solution for now

Have the player's `death` counter increase, without increasing their `kills` counter.

ezgif-1-60f4ca8066b5

ahv15:
### Contains
"Fixes #145"
Updates the death counter every time the player dies due to fall damage.

### How to test
In game give fall damage to the player until the player dies.

- [x] Update the death counter due to fall damage
- [ ] Teleport back to the base on death

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
@@ -72,7 +72,9 @@ public class PlayerDeathSystem extends BaseComponentSystem {
         if (player.hasComponent(PlayerCharacterComponent.class)) {
             event.consume();
             String team = player.getComponent(LASTeamComponent.class).team;
-            updateStatistics(event.getInstigator(), "kills");
+            if(event.getInstigator().getComponent(PlayerStatisticsComponent.class)!=null) {
+                updateStatistics(event.getInstigator(), "kills");
+            }
             updateStatistics(player, "deaths");
             dropItemsFromInventory(player);
             player.send(new RestoreFullHealthEvent(player));


Cervator (changes requested):
Trivial style fix needed, and as noted in the description the functionality isn't quite complete yet but I can confirm a "death" was counted :-)

However, I wonder if this is the right approach 🤔 Is this not merely adding a count for the deaths when the player _would_ have died, but doesn't, as opposed to making the player actually die properly to fall damage, which would automatically involve the respawn on the flag base just as if killed by another player? And maybe even increase the death counter through that logic too.

ahv15:
All I did was make sure that if the player died and was not killed by anyone there would be no null pointer exception, which before blocked the flow, but now allows for the code to continue as can be seen by restoration of health. But the problem lies in teleportation back to the base. Maybe I didnt really understand what you meant?

Cervator:
Hmm, on looking at the code more closely, you're right - I thought the death would have been handled elsewhere, and this just handled the counter. But that method seems to approximate player death _without_ actually dying - as you say it includes the teleport step and everything, that just doesn't work for some reason.

That helps - sorry for my misunderstanding, I haven't dug deep in this part of the code lately :-)

As for why the teleport doesn't work - hmm. The flow probably still differs - usually when a player kills another player it is handled somewhere with an `instigator` entity to quiz for the killer's details and the damage type that finished off the victim. I don't see that on this event handler, so maybe there's another place involved when it is a player kill, which could prepare some variables that lets the teleport work? 🤔 

That's me just thinking out loud more than anything, would need to look around some more, but it is nearing in on midnight here, I should probably get some sleep  ... 😅

ahv15:
ok sure , Just wanted to mention that when I had a look at it I noticed that a few components which had been added such as the location component , rigid body component etc ,was removed for some reason which probably should not be happening , but right when I thought I found the error the behavior started changing completely so hopefully I will be able to observe something more consistent,  wont disturb your sleep anymore :grinning:.

stefaniamak:
Why was that closed? 🤔 @ahv15

jdrueckert:
I believe because of
> However, I wonder if this is the right approach 🤔 Is this not merely adding a count for the deaths when the player would have died, but doesn't, as opposed to making the player actually die properly to fall damage, which would automatically involve the respawn on the flag base just as if killed by another player? And maybe even increase the death counter through that logic too.

ahv15:
@stefaniamak The PR I made only updated the statistics but for some reason death due to fall damage did not trigger teleport back to the base. As I did not take a further look at it I decided to close it and open it back later if I try to fix this later on.

ahv15:
I guess something is going on in some other class which prevents the teleport

stefaniamak:
I see, so the flag didn't teleport too. I was to suggest merging just the counter and having the issue only being at teleporting the player, but that could be really confusing to the player when seeing their score. Alright, got it 👍

PR #160, opened 2020-09-04
Allows all players to restart the game after 10 seconds

### LINKED ISSUE #154: Logic around Restart button at Game Over Screen
### Current Situation

The only one who has the permission to restart the game is the Host of the game. If you are a host, you get an extra button to appear at your Game Over Screen, which allows you to restart the match.

restart button

### Desired Situation

Provide restart permission to all players, by changing the current logic of the Restart Button. It should be taken into consideration factors for instance:
- not being able to restart right away, in case some players want to have a closer look at the stats

ahv15:
### Contains
"Fixes #154"
Allows all the players to restart the game. The restart button appears after 10 seconds allowing the players to check their stats in the mean time.

### How to test
Host and Join a multiplayer game using 2 or more host/clients and once the game is completed everybody should be able to restart the game.

- [x] Allow everybody to restart the game.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
@@ -73,7 +73,6 @@ public final class LASUtils {
     public static final String DEATH_SCREEN = "engine:DeathScreen";
     public static final String ONLINE_PLAYERS_OVERLAY = "engine:onlinePlayersOverlay";
 
-    public static final String RESTART_PERMISSION = "restart";
     public static final String DROPPED_FLAG = "dropped_flag_indicator";
     public static final long FLAG_TELEPORT_DELAY = 20000;
 
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
@@ -15,6 +15,10 @@
  */
 package org.terasology.ligthandshadow.componentsystem.controllers;
 
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.event.ReceiveEvent;
@@ -26,7 +30,6 @@ import org.terasology.ligthandshadow.componentsystem.components.LASTeamComponent
 import org.terasology.ligthandshadow.componentsystem.components.PlayerStatisticsComponent;
 import org.terasology.ligthandshadow.componentsystem.events.GameOverEvent;
 import org.terasology.ligthandshadow.componentsystem.events.RestartRequestEvent;
-import org.terasology.logic.permission.PermissionManager;
 import org.terasology.logic.players.LocalPlayer;
 import org.terasology.logic.players.PlayerCharacterComponent;
 import org.terasology.logic.players.PlayerUtil;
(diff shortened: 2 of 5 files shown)


stefaniamak:
I hosted a game and joined with 2 more clients at it, but I don't seem to have the button appear 🤔 (I gave them more than 10s to appear)

image

stefaniamak (commented):
I hosted a game and joined with 2 more clients at it, but I don't seem to have the button appear 🤔 I gave them more than 10s to appear, and also restarted and finished the game a second time.

image

stefaniamak:
Also, it would be cool to have a timer which would indicate when the button will be active! But that is deffinitelly for a next PR.

ahv15:
This is very weird the restart button should not be visible for 10 seconds even for the host

stefaniamak:
> 
> 
> This is very weird the restart button should not be visible for 10 seconds even for the host

huh 🤔 I did had it right away for the host. Was there maybe any other step to test this other than joining with local clients to a game?

ahv15:
It works fine when I run it from my workspace
Screenshot (410)
Screenshot (411)
This is how it looks for me just checked it out again.

Cervator (changes requested):
Can confirm this _partially_ works :-)

I think the quirks about the button might just be from stale build files - I've seen that happen a fair bit in modules lately. I did a rebuild all in IntelliJ (and I bet `gradlew jar game` would do it too) then was able to see the button correctly go visible after 10 seconds.

The _next_ part is what doesn't work: only if the reset button is clicked by the _host_ (the authority) does the score actually correctly reset. Otherwise (which would always be the case in headless multiplayer) the score doesn't reset so the next point for the winning side wins it again :-)

The reason for this is likely due to `RestartSystem.onRestartRequest` being set to trigger only on the authority side on the network, meaning even though the system itself being registered to `ALWAYS` will run on _both_ the authority and all clients that one event handler will only run correctly on a hosting player. Or it will only correctly trigger via `ClientGameOverSystem` if it resides in the same entity system as the authority. Not entirely sure which.

Seeing the `netFilter` set like that seems kinda rare to me. Much more commonly there'll be a split between client and authority _systems_, with the logic divided between both. So button handling would go into a client system (`ClientGameOverSystem`) that only runs where you've got a player (meaning regular clients and hosting clients, but not headless servers) then the button click will sent some sort of event solely _to_ the authority, which then handles the reset and sends a broadcast to the connected clients

Something like that, anyway There should be some examples of that floating around, even inside LaS, like when a point is scored. https://github.com/MovingBlocks/Terasology/wiki/Entities%2C-Components-and-Events-on-the-Network may also help.

stefaniamak:
Note: there is a `hasRestartPermission` Boolean variable at `GameOverEvent.java`, could you consider removing it as well if it is not needed elsewhere?

stefaniamak:
I've committed the removal of needed permition here: https://github.com/Terasology/LightAndShadow/blob/develop/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java#L105-L110
``` java
if (permissionManager.hasPermission(clientComponent.clientInfo, LASUtils.RESTART_PERMISSION)) {
     redScore = 0;
     blackScore = 0;
     sendEventToClients(new ScoreUpdateFromServerEvent(LASUtils.RED_TEAM, redScore));
     sendEventToClients(new ScoreUpdateFromServerEvent(LASUtils.BLACK_TEAM, blackScore));
}
```

Please check whatever is yet related to that permission and remove it, considering that it is unused code after this PR's changes. I mentioned a place there is a related variable at my above comment, then I would suggest checking some events (e.g. `RestartRequestEvent`) and remove what is now unnecessary 👍 

The removal of the unused code might not be that necessary for this PR to be merged (can possibly be done within a later of PR), but I think it would be best to be done now, as there are no blockings on removing that code now :) If you _don't_ remove it now and this PR gets merged, I would suggest you create an issue addressing this be done.

stefaniamak (dismissed):
I've committed the removal of the permition from this file: https://github.com/Terasology/LightAndShadow/blob/develop/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java#L105-L110
``` java
if (permissionManager.hasPermission(clientComponent.clientInfo, LASUtils.RESTART_PERMISSION)) {
     redScore = 0;
     blackScore = 0;
     sendEventToClients(new ScoreUpdateFromServerEvent(LASUtils.RED_TEAM, redScore));
     sendEventToClients(new ScoreUpdateFromServerEvent(LASUtils.BLACK_TEAM, blackScore));
}
```

Please check whatever is yet related to that permission and remove it, considering that it is unused code after this PR's changes. I mentioned a place there is a related variable at my above comment, then I would suggest checking some events (e.g. `RestartRequestEvent`) and remove what is now unnecessary 👍 

The removal of the unused code might not be that necessary for this PR to be merged (can possibly be done within a later of PR), but I think it would be best to be done now, as there are no blockings on removing that code now :) If you _don't_ remove it now and this PR gets merged, I would suggest you create an issue addressing this be done.

Cervator (approved):

Cervator:
Can confirm the tweak from @stefaniamak did the final trick and collapsed the quantum code state into one that works all the time, not just all the time for _some_ and none of the time for others 😁 

The last few bits of the restart permission were trivial to remove so I did it real quick and added it to a commit here. 5 total commits with 4 contributors, now that's team work. Well done everybody and thank you! Realistically replayable LaS in headless multiplayer FTW!

PR #161, opened 2020-09-10
Add a timer to indicate when the restart button gets enabled

ahv15:
### Contains
Adds a timer to show when the the restart button will be enabled.

### How to test
In addition to this PR check Terasology/LightAndShadowResources#60.
Join a single player or multiplayer game and when the game is over a timer should appear to show when the restart button will be enabled.

- [x] Timer to show when restart button will be enabled.

### DIFF
--- a/overrides/engine/ui/ingame/deathScreen.ui
+++ b/overrides/engine/ui/ingame/deathScreen.ui
@@ -336,6 +336,26 @@
               }
             },
             {
+              "type": "UILabel",
+              "text": "Timer",
+              "family": "noBackground",
+              "id": "timer",
+              "layoutInfo": {
+                "height": 30,
+                "width": 45,
+                "position-top": {
+                  "target": "BOTTOM",
+                  "widget": "detailsUIBox",
+                  "offset": 10
+                },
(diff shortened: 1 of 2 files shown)


ahv15:
There probably will be some style issues as I reset my workspace, hopefully checkstyle is working fine. Need to add an additional PR to LightAndShadowResources too, to add the disabled mode for the UI Button in the client game over skin.

stefaniamak (changes requested):
Works a supposed to both host and client 👍 

I would only suggest you have the timer disappear when reaching `0`. What I mean: have the timer reach `1` and after that disappear while the button gets enabled (`0` doesn't appear at all).

_Also_: please note at the PR description that you have to check https://github.com/Terasology/LightAndShadowResources/pull/60 before testing. (You can add it at the "How to test" section)

ahv15:
@stefaniamak  made the change and it disappears, did try setting the timer as invisible but the timer remained invisible the entire time even if I set it as visible or invisible so I tried it out in this way instead.

stefaniamak (approved):
Nice! It works great 👍

PR #168, opened 2020-11-14
feat: remove minimap

jdrueckert:
The current LaS arena is very small and provides a limited set of strategic possibilities to prevent the opposing team from getting to the flag and bring it back to their base.
One of the strategic possibilities is to obfuscate where the own team members are located for instance by hiding behind generated world elements such as trees or by building barricades.
With `Minimap` in play, this strategy is not possible as the location of players is always visible. 

This is why I propose to remove `Minimap` from `LightAndShadow`.
Once we create additional play arenas in which `Minimap` would be beneficial again, it can easily be re-integrated. Ideally, we could then make the use of `Minimap` a configurable setting, so that players can decide themselves whether they want to play with it or without. Currently, this alternative is not feasible as the gameplay configuration possibilities are limited if not broken.

### DIFF
--- a/module.txt
+++ b/module.txt
@@ -16,7 +16,6 @@
         { "id": "Inventory", "minVersion": "1.1.0" },
         { "id": "ItemRendering", "minVersion": "1.1.0" },
         { "id": "LightAndShadowResources", "minVersion": "1.0.0" },
-        { "id": "Minimap", "minVersion": "1.0.0" },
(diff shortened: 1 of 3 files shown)


skaldarnar (approved):

jdrueckert:
[Remove Minimap](https://trello.com/c/fXJoECmf/40-remove-minimap)

PR #170, opened 2020-11-14
feat(JOML): migrate card system

pollend:
This is a straight forward migration to JOML. still not sure if the block family change is correct.


jdrueckert:
Just found out that even `door` items are currently not placable, so we need to investigate this separately anyway.
With this PR cards are still properly placable, so I believe it's good as it is.

jdrueckert (approved):

jdrueckert (approved):

PR #184, opened 2021-01-09
chore: remove EnemySystem

pollend:
from what I can tell I don't think this is used at all. just wondering what the plan was for this @jdrueckert.


skaldarnar (approved):

skaldarnar:
As this is about enemies _and_ it is from synopia I suspect something along the lines of "behavior trees v2" and different vision for where LaS should go at that point in time. I think we weren't that clear on the arena shooter multi-player back then...

PR #193, opened 2021-03-09
Randomize player respawn location

### LINKED ISSUE #85: Randomize player respawn location
Follow up to #82 

In https://github.com/Terasology/LightAndShadow/blob/master/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java#L45

> Ideally, we should not be using the exact location of the base. That would mean when multiple players die, they always land on top of each other. Since the base is big enough, we could teleport to a random position on the base as in LASUtils.getTeleportDestination(team)) + Vec3(random(-1, 1), random(-1, 1), 0).

> Again, here make sure that the players always face the opponent team's base. Maybe extract the above completely into a method and use it in the team selection logic as well.

ahv15:
### Contains
"Fixes #85"
Respawn the player at random locations on the base so that teamates don't spawn on top of each other.

### How to test
In  singleplayer mode in console type kill to kill the player multiple times, he should respawn at different positions each of those times. 

- [x] Randomize player respawn.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
@@ -15,6 +15,8 @@
  */
 package org.terasology.ligthandshadow.componentsystem.controllers;
 
+import java.util.Random;
+
 import org.joml.Vector3f;
 import org.joml.Vector3fc;
 import org.terasology.assets.management.AssetManager;
@@ -58,6 +60,8 @@ public class PlayerDeathSystem extends BaseComponentSystem {
     @In
     private BlockManager blockManager;
 
+    private Random random = new Random();
+
     /**
      * Empty the inventory and send player player back to its base with refilled health.
      * This is a high priority method, hence it receives the event first and consumes it.
@@ -77,7 +81,8 @@ public class PlayerDeathSystem extends BaseComponentSystem {
             updateStatistics(player, "deaths");
             dropItemsFromInventory(player);
             player.send(new RestoreFullHealthEvent(player));
-            player.send(new CharacterTeleportEvent(LASUtils.getTeleportDestination(team)));
+            Vector3f randomVector = new Vector3f(-2 + random.nextFloat()*4, -2 + random.nextFloat()*4,0);
+            player.send(new CharacterTeleportEvent(randomVector.add(LASUtils.getTeleportDestination(team))));
         }
     }
 
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
@@ -81,7 +81,7 @@ public class PlayerDeathSystem extends BaseComponentSystem {
             updateStatistics(player, "deaths");
             dropItemsFromInventory(player);
             player.send(new RestoreFullHealthEvent(player));
-            Vector3f randomVector = new Vector3f(-2 + random.nextFloat()*4, -2 + random.nextFloat()*4,0);
+            Vector3f randomVector = new Vector3f(-2 + random.nextInt(5), 0, -2 + random.nextInt(5));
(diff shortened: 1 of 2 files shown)


DarkWeird:
This works at death only?
What about initially spawn location?

ahv15:
I was going to do that too but as the issue was regarding only respawn events I did not do so. Let me add a quick commit to fix that too.

ahv15:
I wasn't sure whether it would be ok to solve separate issues in one PR.

DarkWeird:
> I wasn't sure whether it would be ok to solve separate issues in one PR.

I see both issues as one problem: "Players spawning at one position in LaS"
But ok.

feel free to  do what you doing :p (idk how LaS works )

ahv15:
It is the same problem but I assumed there might be some sort of reason randomize 'respawn' was specified thats it.

ahv15:
@skaldarnar is it ok if I get the player approximately facing the enemies base, as what I plan on doing is use my player and approximately face the opponents base and get the corresponding Quaternion for that orientation and use that Quaternion to set orientation/rotation every time player spawns or respawns?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -77,7 +81,8 @@ public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterC
             updateStatistics(player, "deaths");
             dropItemsFromInventory(player);
             player.send(new RestoreFullHealthEvent(player));
-            player.send(new CharacterTeleportEvent(LASUtils.getTeleportDestination(team)));
+            Vector3f randomVector = new Vector3f(-2 + random.nextFloat()*4, -2 + random.nextFloat()*4,0);
```
skaldarnar: Instead of doing `random.nextFloat()*4` which generates a random number `x ∈ [0..4]` you can do `random.nextInt(5)` which generates a random number `n ∈ [0..5)` (exclusive upper bound).

skaldarnar: Oh, and Terasology uses the Y-axis for up/down, but I think you want to randomize the position in the x/z plane. Thus, the `0` should be the second argument.

```java
            Vector3f randomVector = new Vector3f(-2 + random.nextFloat()*4, 0, -2 + random.nextFloat()*4);
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -65,7 +70,8 @@ private String setPlayerTeamToTeleporterTeam(EntityRef player, EntityRef telepor
     }
 
     private void handlePlayerTeleport(EntityRef player, String team) {
-        player.send(new CharacterTeleportEvent(LASUtils.getTeleportDestination(team)));
+        Vector3f randomVector = new Vector3f(-2 + random.nextFloat()*4, -2 + random.nextFloat()*4,0);
```
skaldarnar: Same remarks as above apply.

skaldarnar (changes requested):
This PR needs to be updated against the latest changes in engine/Omega. I think the only conflict here should be that things moved into an `engine` package:

`org.terasology.entitysystem...` >>> `org.terasology.engine.entitysystem`

ahv15:
Will make the changes.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Just wanted to mention that I am considering the range [-2,2] for the random vector even though the actual range should be [-2.5,2.5] as I didn't want players spawning on the edge of the base.

ahv15:
I am tempted to send in a new PR as I messed up my workspace. I keep forgetting to update the workspace.

jdrueckert:
Superseded by https://github.com/Terasology/LightAndShadow/pull/197

PR #197, opened 2021-03-25
feat: Randomize player respawn location

### LINKED ISSUE #85: Randomize player respawn location

ahv15:
### Contains
"Fixes #85"
Respawn the player at random locations on the base so that teamates don't spawn on top of each other.

### How to test
In  singleplayer mode in console type kill to kill the player multiple times, he should respawn at different positions each of those times. 

- [x] Randomize player respawn.


ahv15:
The ranges were modified on purpose as the player spawns outside the base at times.

skaldarnar (approved):

PR #198, opened 2021-04-04
stop trees from spawning on team bases by adding LaSTreeProvider and LaSExtraPositionFilters

maxjwebber:


src/main/java/org/terasology/las/LaSExtraPositionFilters.java
```diff
@@ -0,0 +1,77 @@
...
+                                (myx == redx + xOffset && myz-1 == redz + zOffset)||
+                                (myx == blackx + xOffset && myz+1 == blackz + zOffset)||
+                                (myx == redx + xOffset && myz+1 == redz + zOffset))
+
+                            return false;
+                    }
+                }
+                return true;
+            }
+        };
```
pollend, line 74: There isn't really a need to have this separated out from your implementation. Better to do this type of separate after the fact when you are using the same logic throughout the project. Makes it hard to understand what is happening when the details are separated out into another utility class.

pollend (changes requested):

src/main/java/org/terasology/las/LaSExtraPositionFilters.java
maxjwebber: > 
> 
> There isn't really a need to have this separated out from your implementation. Better to do this type of separate after the fact when you are using the same logic throughout the project. Makes it hard to understand what is happening when the details are separated out into another utility class.

Hello, and thanks very much for your response and review.

There may be a benefit to adding more position filters, specific to Light&Shadow, rather than declaring them anonymously in the Tree Provider. Another issue I'm working on resolving is that some trees penetrate the Floating Platform when they spawn under it. This could be resolved by adding a filter which restricts trees from spawning under the platform.
 
This version of LaSTreeProvider uses copied class code from DefaultTreeProvider which I edited to add the new filter. I would also ask that you look at [my previous commit ](https://github.com/Terasology/LightAndShadow/pull/198/commits/8a4410394eef6274f86854d086a271a4dcc31b09) in which LaSTreeProvider inherits from DefaultTreeProvider and overrides the getFilters() method. While this is much more readable, the game would not build. Talking to other developers on Discord, it was suggested that this may have been due to a permission/whitelist issue that did not come up when I copied the class code.

What do you think?

maxjwebber:
I have renamed my branch to reflect changes to my methodology, which resulted in this pull request being closed. I will open a new one once I've made changes.

pollend:
@maxjwebber might still want to look into this? why did you delete it?

maxjwebber:
> 
> 
> @maxjwebber might still want to look into this? why did you delete it?

Didn't realize changing a branch name would close related pull requests. I changed the name back and reopened the request.

skaldarnar:
Not sure where I left this commment (probably some GSOC proposal), but I don't think we need a custom tree provider here. Instead, it should be possible to make use of the fact that trees are only generated on the surface, and therefore modifying the surface facet should do the trick. Lakes do this to avoid that grass spawns on the water.

That said, I give you credit for the abstraction to allow for additional filters (to expand it later on) :nerd_face:

jdrueckert:
The issue this was attempting to fix was fixed by #216

PR #201, opened 2021-04-09
feat: A basic implementation for an npc

ahv15:
### Contains
Uses a system for identifying enemies (similar to MR) which can be used by npc's (maybe defence mechanisms and traps in the future) to identify and attack enemies.

### How to test
Create a multiplayer game with players on both teams and then spawn a pawn near an enemy. Make sure to add the behavior and LaSTeam component to the pawn's prefab(will be sending a separate PR once this PR is finalized).

- [x] Identify enemies
- [ ] Npc's that attack their enemy

### DIFF
--- a/assets/behaviors/pawn.behavior
+++ b/assets/behaviors/pawn.behavior
@@ -1,0 +1,25 @@
+{
+    dynamic: [
+        {
+            guard: {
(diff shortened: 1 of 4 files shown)


ahv15:
Will be moving pawn.behavior to LaSResources once the PR has been finalized.

PR #202, opened 2021-04-12
Raise Floating Platform so trees cannot penetrate/collide with it.

maxjwebber:
image
The above image illustrates a bug present where trees can sometimes grow tall enough to penetrate/collide with the floating platform.

MountainProvider can elevate terrain up to y= 50 units. The tallest tree, the birch, is 35 units tall. Currently the floating platform spawns at y=60. By setting the floating platform height to 90, the floating platform is guaranteed to be higher than the tallest tree at the highest elevation.


pollend:
ah clever.

pollend (approved):

skaldarnar:
This will be addressed by #216 (which does not mean that this PR was not valid - we just forgot to merge it, it seems :see_no_evil: )

PR #203, opened 2021-04-17
feat: Prevent players from joining the larger team

ahv15:
### Contains
If a player tries to join the larger team, the player is notified through the chat box to join the other team. Requires https://github.com/Terasology/LightAndShadowResources/pull/63 to test this.

### How to test
In  multiplayer try to join the larger team.

- [x] Prevent players from joining the larger team.


jdrueckert:
[Improve LaS team balancing](https://trello.com/c/kawWZtVY/65-improve-las-team-balancing)

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +46,35 @@
...
-        String team = setPlayerTeamToTeleporterTeam(player, entity);
-        handlePlayerTeleport(player, team);
+        if (properTeamSize(entity, player)) {
+            String team = setPlayerTeamToTeleporterTeam(player, entity);
+            handlePlayerTeleport(player, team);
+        }
+    }
+
+    private boolean properTeamSize(EntityRef teleporter, EntityRef player) {
+        int maxTeamSizeDifference = 3;
```
jdrueckert: Ideally, this should be a constant, even more ideally a "world-configurable" one as in when you start to host an LaS game, you can configure it and then it applies for everyone.
World configuration is currently being fixed though, so I'm not sure how well you can currently test that - but the general idea should be informed about in the world gen tutorial IIRC.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +46,35 @@
...
         EntityRef player = event.getInstigator();
-        String team = setPlayerTeamToTeleporterTeam(player, entity);
-        handlePlayerTeleport(player, team);
+        if (properTeamSize(entity, player)) {
+            String team = setPlayerTeamToTeleporterTeam(player, entity);
+            handlePlayerTeleport(player, team);
+        }
+    }
+
+    private boolean properTeamSize(EntityRef teleporter, EntityRef player) {
```
jdrueckert: It's recommended to prefix boolean variables and functions with "is", so that it's clear what the respective `true` and `false` case will indicate.

```suggestion
    private boolean isProperTeamSize(EntityRef teleporter, EntityRef player) {
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +46,35 @@
...
+        int maxTeamSizeDifference = 3;
+        int oppositeTeamCount = 0;
+        int teleporterTeamCount = 0;
```
jdrueckert, line 80: IMO we should track this more globally instead of counting again every time a player tries to join a team.

jdrueckert (changes requested):
As a general note, I don't think we should prevent a player completely from joining a team that already has more players than the other (respecting the configured diff).
Reasoning behind this: From CTF rounds I played with friends I recall situations such as
* some people being far more experienced and skilled, thus playing alone against multiple "beginners"
* people agreeing on specific team constellations before-hand and for instance due to loading times, most people of one team being "online" before the people of the other team. While it's not as big a problem as we have our "lobby" platform, I still think it's bad UX if those players need to wait for the others to join the game and enough people to teleport into the arena before they can teleport into the arena themselves.

However, if for instance you play with random people, you might want to _really_ enforce this, which makes me think this should be configurable by the person hosting the game / maintaining the server.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: @jdrueckert I don't remember seeing this there, but would it be ok if we could configure this through the settings, maybe under the player configuration part. If there is any better alternative could you point it to me, I will work on that.

ahv15: Did you mean configuring it through the universe setup. My game seems to crash when I try to configure the world.

jdrueckert: Yes I meant configuring through the universe setup.
Huh, I'm also getting a `java.util.concurrent.ExecutionException: java.lang.ExceptionInInitializerError` :thinking: For `CoreGameplay` it works...
I recall there _was_ an issue at some point due to pre-generation stage in LaS trying to access something that's only loaded later (IIRC that came from `StaticCities`). However, I'm rather sure we fixed that :thinking: 

@ahv15 can you open up an issue for this?

jdrueckert: On second thought, the universe setup is only for world-gen stuff, so that's not really suitable for gameplay configuration :thinking:
The least we should do as a first step is create constants that allow for a central "configuration". As a follow up (writing up an issue and resolving it in another PR) we can consider something like an in-game configuration mechanism for the host, such as a screen, some "buttons" or an additional dialogue option with the fool or something like that.

ahv15: Will do.

ahv15: An additional dialogue option might be preferable in this case as currently there aren't many configurable settings for LaS and adding a single screen/button seems unnecessary. Another option might be to make use of the settings screen to do this and allow all gameplay configuration over there. Maybe notify the player that gameplay configuration can be done through the magic fools dialogue.
Maybe under the player settings in autoconfig
image
Or an additional option here to configure gameplay specific things
image

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Done

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Done

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -45,7 +33,11 @@
     @In
     EntityManager entityManager;
 
-    private Random random = new Random();
+    private final Random random = new Random();
+
+    private final int maxTeamSizeDifference = 3;
+    private int redTeamCount;
+    private int blackTeamCount;
```
skaldarnar: This is state that should be kept in a component - I think there is some game state entity that holds information relevant to a "game" of LaS.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -45,7 +33,11 @@
     @In
     EntityManager entityManager;
 
-    private Random random = new Random();
+    private final Random random = new Random();
+
+    private final int maxTeamSizeDifference = 3;
```
skaldarnar: This is a _configuration_ that might as well come from a component, and thus could be modified/set by a prefab.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
jdrueckert: For this you'd need to register settings based on the gameplay modules being active.
While this is generally an interesting and much needed topic that would also benefit the "Input" settings, I'm not sure how feasible it is at the moment and which steps exactly I'd entail.
Feel free to research this as a potential follow-up, though :+1: (and consider writing the results of your research up as a follow-up issue)

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Actually game state information is scattered throughout the code as variables (probably should keep track of team scores through the ScoreComponent rather than variables) and components which are associated with separate entities . So for the time being I could create a component for teamCount and attach it to a gameState prefab. Maybe later on move everything to the gameState prefab which will keep track of all the game information if that is preferred.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Regarding this I am actually looking into creating a GameplayConfiguration file which makes use of AutoConfig and will keep track of the maxTeamSizeDifference and any future configurations required with respect to LaS. As some of the configurations make use of it , the required UI is automatically created and the UI will be generated only when LaS is active this should be useful. This is something I didn't really look into or test out properly so it might not work the way I expected it to.  So I am planning on taking a look at that too before I decide if I make use of a Config file or convert it into a component.

ahv15: This actually requires a lot of additional setup and was not as concise as I hoped it would be.  So I mostly will be creating a component for this too and making use of commands to set the maxTeamSizeDifference. The main reason I was hoping to use AutoConfig was because it would be concise but as that is not the case it might not be the best option.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: While looking at some suitable alternatives I did see some issues, so for the time being I will just add a command through which we can set the maxTeamSizeDifference and once the list of configurations becomes larger maybe create a separate screen then.

ahv15:
@skaldarnar If these changes seem reasonable I will send in a separate PR to LightAndShadowResources too.

jdrueckert (changes requested):
Very sorry for the delay... @skaldarnar and I had a bit of a discussion about counting the team members every time vs keeping state.
Coming from a database background I often tend to go for storing and updating information to look up when needed rather than to compute it on-demand.
We agreed that keeping state does have its place, especially when doing something very frequently. However, one of his arguments that I do have to agree with, is that in the case of players joining a team - which is happening a few times at the beginning of a game only - it might actually be easier and not too costly to count every time.
Especially as with this we do not have to care about people disconnecting from the game etc.

Regarding the diff component... IMO a central `LASConfig` component would make more sense than multiple small ones. I actually would've thought we already have an artificial entity that's for instance also tracking the score, but I couldn't find for instance the `ScoreComponent` in an entity dump in the first place, so... maybe not? Anyway, every aspect of the game should ideally be present in some way in our Entity-Component structure, so having a gameplay entity that's keeping the configuration and central state like the score that should not be tracked by individual player entities would make sense.

TL;DR let's go for counting every time a player tries to join a team after all (and apologies for the back and forth) and please instead of a small `MaxTeamSizeDifferenceComponent` go for a bigger `LASConfigComponent` instead. Also, instead of accessing a config prefab directly, we should go for a gameplay entity that has the config component and read / update that component on that entity.

ahv15:
Ok will make those changes (didn't see the problem with the disconnecting players). Is there any particular scenario where it would be preferred to use prefabs directly? Also are there any plans to move game state information to this entity. I was going to add some more configurable items (friendly fire, goal score etc) once this PR is finalized before creating a screen for configuration but due to my final exams next week I probably can't spend too much time working on it.

skaldarnar:
>  Is there any particular scenario where it would be preferred to use prefabs directly?

Not sure whether I understand the question correctly. A prefab is just a blue print for constructing an entity and pre-fill it with some default values. 
Given that, I'd think that using a prefab for the default configuration for Light and Shadow would be a good idea.

ahv15:
> > Is there any particular scenario where it would be preferred to use prefabs directly?
> 
> Not sure whether I understand the question correctly. A prefab is just a blue print for constructing an entity and pre-fill it with some default values.
> Given that, I'd think that using a prefab for the default configuration for Light and Shadow would be a good idea.

@skaldarnar I had this doubt because @jdrueckert suggested using a gameplay entity rather than a gameplay prefab for configuration. Just wanted to know when which is preferred.

jdrueckert:
>@skaldarnar I had this doubt because @jdrueckert suggested using a gameplay entity rather than a gameplay prefab for configuration. Just wanted to know when which is preferred.

@ahv15 I was targeting exactly the
> A prefab is just a blue print for constructing an entity and pre-fill it with some default values.

aspect. In-game, we want to use the entity that was pre-filled using a prefab instead of directly accessing the prefab.

Will have another look hopefully tomorrow, today still busy with core efforts :see_no_evil:

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +62,49 @@
...
+        Iterable<EntityRef> characters = entityManager.getEntitiesWith(PlayerCharacterComponent.class,
+                LASTeamComponent.class);
+
+        for (EntityRef character : characters) {
+            String otherPlayerTeam = character.getComponent(LASTeamComponent.class).team;
+            if (teleporterTeam.equals(otherPlayerTeam)) {
+                teleporterTeamCount++;
+            } else if (!otherPlayerTeam.equals(LASUtils.WHITE_TEAM)) {
+                oppositeTeamCount++;
+            }
```
jdrueckert, line 92: Afaik there's currently no possibility to join the white team (and the white team is actually on of the aspects of the gameplay that's targeting at the original gameplay idea, but is not relevant to the current state of LightAndShadow. In the context of cutting things down, I'd say, we should rather remove it from LaS instead of further integrating it in logic we clean up or build

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +62,49 @@
...
+        }
+        if (teleporterTeamCount - oppositeTeamCount < maxTeamSizeDifference) {
+            teleporterTeamCount++;
+            gameEntity.getComponent(GameStateComponent.class).redTeamCount = teleporterTeam.equals(LASUtils.RED_TEAM) ? teleporterTeamCount : oppositeTeamCount;
+            gameEntity.getComponent(GameStateComponent.class).blackTeamCount = teleporterTeam.equals(LASUtils.RED_TEAM) ? oppositeTeamCount : teleporterTeamCount;
+            return true;
+        } else {
+            gameEntity.getComponent(GameStateComponent.class).redTeamCount = teleporterTeam.equals(LASUtils.RED_TEAM) ? teleporterTeamCount : oppositeTeamCount;
+            gameEntity.getComponent(GameStateComponent.class).blackTeamCount = teleporterTeam.equals(LASUtils.RED_TEAM) ? oppositeTeamCount : teleporterTeamCount;
```
jdrueckert: why do you need to change the component in this case?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +62,49 @@
...
+            player.getOwner().send(new ChatMessageEvent("The " + teleporterTeam + " team has more players so please join the " + LASUtils.getOppositionTeam(teleporterTeam)
```
jdrueckert: Nitpick: This formulation can be a bit confusing as depending on `maxTeamSizeDifference` we may not directly "complain" once one team has more players, but only after it has a number of players more than the other.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +62,49 @@
...
+            return false;
+        }
+    }
+
+    private void setGamePrefab() {
+        if (entityManager.getCountOfEntitiesWith(LASConfigComponent.class) == 0) {
+            gameEntity = entityManager.create("LightAndShadowResources:gameEntity");
+        } else {
+            gameEntity = entityManager.getEntitiesWith(LASConfigComponent.class).iterator().next();
+        }
```
jdrueckert: I don't understand why we need this :thinking:

jdrueckert (changes requested):

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: This isn't needed right now, but in the future if the gameEntity is created elsewhere for configuration or game state this would be needed then.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Will change that

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: I was wondering if it would be necessary to store the team count, as in the case of disconnecting players the stored information would be incorrect. There would have to be some way to update the information regularly or whenever a player disconnects. This is kind of unnecessary but I added it so that the updates could be made more frequently. Should I remove the team count information altogether?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Yeah currently the white team only helps identify the lobby players, so if that's not needed should I remove all instances of it?

jdrueckert: Yes I'd say so

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
jdrueckert: As we're back to counting every time a player tries to join and don't re-use this for that purpose (and also for no other purpose at the moment), I'd say yes, remove it again.
If it comes up in the future again, I'd assume it's pretty easy to reintroduce.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
jdrueckert: Then I'd say, reintroduce it when it's actually needed.
We have a history of introducing things that might be useful or needed later on but never avtually getting to that point and thus cluttering our codebase.
Btw, that's not to say I don't believe in you getting to that point, just that with that track record we learned it's better to not do it irrespective of personal beliefs and hopes.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: @jdrueckert  Actually the white team is necessary to differentiate the lobby players from the players with teams, as the ClientSkinSystem initializes the lobby players with different HUD's and appearances when compared to those with teams. So shouldn't the white team be left alone?

jdrueckert: Where did you find that?

ahv15: image
image
Multiplayer is glitching a bit for me right now, have to clean my workspace up before I can confirm but in ```ClientSkinSystem.java``` , ```onCreateDefaultVisualCharacter``` initializes the ```skeletalMeshComponent``` based on the initial team of the player and ```onAwaitedLocalCharacterSpawnEvent``` uses ```setHealthHUD``` to initialize the healthBar based on the initial team of the player, so lobby players should look different and have different health bars.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -57,8 +61,47 @@
...
+                        + LASUtils.getOppositionTeam(teleporterTeam) + " team.", EntityRef.NULL));
+            }
+            return false;
+        }
+    }
+
+    private void setGameEntity() {
+        if (entityManager.getCountOfEntitiesWith(LASConfigComponent.class) == 0) {
+            gameEntity = entityManager.create("LightAndShadowResources:gameEntity");
+        }
```
skaldarnar: You've technically build _lazy evaluation_ for the game entity - it is only created on first access. Is that something you did on purpose?

What happens if you load a game of LightAndShadow and there is already a game state entity? 

I would have expected that we reuse the entity if there is one (and probably throw and error if there are two or more, for whatever reason).

```suggestion
		 int numberOfGameStateEntities = entityManager.getCountOfEntitiesWith(LASConfigComponent.class);	
        if (numberOfGameStateEntities == 0) {
            gameEntity = entityManager.create("LightAndShadowResources:gameEntity");
        } else if (numberOfGameStateEntities == 1) {
        	 gameEntity = entityManager.getEntitiesWith(...).get(0) // get the first entity
        } else {
        	// more than two game state entities
        	// log an error or warning and crash the game mode (can we crash back to main window?)
        }
```

Also, we can skip this if `gameEntity` is already set (not `null` and not the `EntityRef.NULL`), can't we?

ahv15: I didn't consider the case of using entities when creating a saved game, I initially did use an entity if it already existed but as the entity wasn't created anywhere else currently, I removed it. Will test this out in saved games and make the change. Will also check whether the entity has been initialized or not.

skaldarnar (changes requested):
We're getting there!

I would like to make one proposal for `setGameEntity()` and how we use it.
The typical use case in your code is to do the following:

```java
        if (gameEntity == null || gameEntity == EntityRef.NULL) {
            setGameEntity();
        }
        // access gameEntity here being sure it is initialized
```

This makes use of the `gameEntity` being a member of the system, but otherwise bears potential errors like forgetting to check for initialization and such. 

If `gameEntity` would not be a local member you usually would call `getGameEntity()` of some other class/system, and you would rely on it being present (and created if needed). 

So, why not apply this pattern here as well? On the caller site, we would have something like this. Note that there are no checks on caller side here - all we express is that we want to have the game entity to work with.

```java
        EntityRef gameEntity = getGameEntity();
```

Within the getter, we combine both the check for initialization and the lazy initialization/caching. A clear benefit is that we only care about this in one location, instead of in many. Plus, it's less code duplication, and will also make it simpler to extract "game entity handling" into a dedicated system (I don't think we should keep the logic of creating the entity in a system like `TeleporterSystem`). The implementation could look like this:

```java
    private EntityRef getGlobalGameEntity() {
        if (globalGameEntity == null || globalGameEntity == EntityRef.NULL) {
            ArrayList<EntityRef> gameEntities =
                    Lists.newArrayList(entityManager.getEntitiesWith(LASConfigComponent.class));

            if (gameEntities.isEmpty()) {
                globalGameEntity = entityManager.create("LightAndShadow:gameEntity");
            } else if (gameEntities.size() == 1) {
                globalGameEntity = gameEntities.get(0);
            } else {
                //TODO: crash hard!
                throw new RuntimeException("Multiple game state entities available.");
            }
        }

        return globalGameEntity;
    }
```

Note, that throwing an exception here currently **does not** crash the game. I've started a [discussion on Discord](https://discord.com/channels/270264625419911192/696830442065625118/847896180742029342) about this. Not sure whether we can/should mitigate this issue here in the system (e.g., choose the first entity, or always the one with lowest entity id or something like that).

ahv15:
> We're getting there!
> 
> I would like to make one proposal for `setGameEntity()` and how we use it.
> The typical use case in your code is to do the following:
> 
> ```java
>         if (gameEntity == null || gameEntity == EntityRef.NULL) {
>             setGameEntity();
>         }
>         // access gameEntity here being sure it is initialized
> ```
> 
> This makes use of the `gameEntity` being a member of the system, but otherwise bears potential errors like forgetting to check for initialization and such.
> 
> If `gameEntity` would not be a local member you usually would call `getGameEntity()` of some other class/system, and you would rely on it being present (and created if needed).
> 
> So, why not apply this pattern here as well? On the caller site, we would have something like this. Note that there are no checks on caller side here - all we express is that we want to have the game entity to work with.
> 
> ```java
>         EntityRef gameEntity = getGameEntity();
> ```
> 
> Within the getter, we combine both the check for initialization and the lazy initialization/caching. A clear benefit is that we only care about this in one location, instead of in many. Plus, it's less code duplication, and will also make it simpler to extract "game entity handling" into a dedicated system (I don't think we should keep the logic of creating the entity in a system like `TeleporterSystem`). The implementation could look like this:
> 
> ```java
>     private EntityRef getGlobalGameEntity() {
>         if (globalGameEntity == null || globalGameEntity == EntityRef.NULL) {
>             ArrayList<EntityRef> gameEntities =
>                     Lists.newArrayList(entityManager.getEntitiesWith(LASConfigComponent.class));
> 
>             if (gameEntities.isEmpty()) {
>                 globalGameEntity = entityManager.create("LightAndShadow:gameEntity");
>             } else if (gameEntities.size() == 1) {
>                 globalGameEntity = gameEntities.get(0);
>             } else {
>                 //TODO: crash hard!
>                 throw new RuntimeException("Multiple game state entities available.");
>             }
>         }
> 
>         return globalGameEntity;
>     }
> ```
> 
> Note, that throwing an exception here currently **does not** crash the game. I've started a [discussion on Discord](https://discord.com/channels/270264625419911192/696830442065625118/847896180742029342) about this. Not sure whether we can/should mitigate this issue here in the system (e.g., choose the first entity, or always the one with lowest entity id or something like that).

Although the latest commit allows the player to join a team, it creates a new entity when a saved game is used so the underlying issue still causes a problem as the initially created entity is not detected using ```getEntitiesWith()```  (when reopening a saved game). So if maxTeamSizeDifference is modified and the game is loaded at a later point of time, then rather than getting the previously configured maxTeamSizeDifference the default value is used instead.

skaldarnar (approved):

jdrueckert (approved):

PR #208, opened 2021-04-29
fix: reintroduce Tasks dep

jdrueckert:
- LaS defines a delta of Tasks:BeaconMark
- as deltas only mention the module of the original asset on the file path, it was overlooked in #207


keturn (approved):
Thank you for catching this!

PR #209, opened 2021-06-10
refactor: merge packages

ahv15:
Merged las and ligthandshadow into a single package based on the below structure.
image
New Structure 
image


jdrueckert:
A few questions/remarks:
1. What do `SetTeamAction` and `ScoreHud` have to do with world gen?
2. In comparison to `barrier` and `flag`, `jobs` and `logic` seem a bit generic
3. What do `ClientSkin`, `Teleporter` and particles have to do with the gamestate?
4. While `PlayerDeath` might result from combat, I'm not sure, it's a very good fit for a `combat` pkg. Same for `PlayerStatistics` which admittedly do consist mostly of K/D but also of captured flags / scored points.

To be honest this feels a bit as if you just copied the contents of one of the two packages into the other without really putting _much_ thought into a suitable structure. Don't get me wrong, splitting up the former `componentsystem` sub-package is a good start, but this definitely does not cover it.

We expect a consistent and meaningful structure on which we can build for the rest of the module overhaul. Naturally, over the course of the overhaul it's possible that there will be some more adjustments to it, but we need a reasonable base to work with.

jdrueckert:
Superseded by https://github.com/Terasology/LightAndShadow/pull/245

PR #210, opened 2021-06-11
refactor: Removing unused code

ahv15:
- [x] Removes unused dialogues and the corresponding action as team selection is no longer done by the magic fool.
- [x] Cleans up instances of the AttackMinionSystem.
- [x] Move CardComponent to LASResources to fix dependency issues.
- [ ] Remove unnecessary components.


src/main/java/org/terasology/las/UI/ScoreHud.java
```diff
@@ -13,24 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.terasology.las.ui;
+package org.terasology.las.UI;
```
skaldarnar: I'd like to stick to the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names) and only use all-lowercase characters for package names.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/LASSystem.java
```diff
@@ -15,46 +15,16 @@
...
-     * Gives player inventory items on game start
-     */
-    @ReceiveEvent
-    public void onPlayerSpawn(OnPlayerSpawnedEvent event, EntityRef player, InventoryComponent inventory) {
-        BlockItemFactory blockFactory = new BlockItemFactory(entityManager);
-        for (int i = 0; i < inventoryManager.getNumSlots(player); i++) {
-            EntityRef itemInSlot = inventoryManager.getItemInSlot(player, i);
-            player.send(new RemoveItemAction(player, itemInSlot, true));
-        }
-    }
```
skaldarnar: This does not look like "unused code". Therefore, please don't remove it in this PR which says **refactor: Removing unused code** :wink: 

You can do this in another PR describing _why_ you want to remove it. If this is about the glitch with the player inventory, it might be worth a try to test out different event handler priorities to ensure that this handler runs last.

src/main/java/org/terasology/ligthandshadow/logic/CardComponent.java
```diff
@@ -1,30 +0,0 @@
...
-package org.terasology.ligthandshadow.logic;
-
-import org.terasology.engine.entitySystem.Component;
-import org.terasology.engine.entitySystem.prefab.Prefab;
-import org.terasology.engine.world.block.family.BlockFamily;
-
-/**
- * This is the component class for playing cards, which are constructed of a top and a bottom block.
- */
-public class CardComponent implements Component {
```
skaldarnar: I know that the cards are currently not used. 

This component and the corresponding system are derived from the logic around doors to turn a single item into a multi-block structure, and treat the blocks as a single union. 

As placing and removing cards does work right now, we could consider to add them to the game rather sooner than later, to allow players building walls with them, for instance. In that case, maybe you only remove the door logic from the system, but leave the rest in-place.

skaldarnar: Oh, and by removing the `CardComponent` in this PR you uncovered that we have a "wrong" dependency between LightAndShadow and LightAndShadowResources. 

The card component _is_ used in there, see `assets/prefabs/cards/*.prefab`. When looking for usage of components in the code base keep in mind that we can omit the `Component` suffix in many cases - here, a search for `"Card":` yields results for the card component.

src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
```diff
@@ -1,148 +0,0 @@
...
-import org.terasology.engine.world.WorldProvider;
-import org.terasology.engine.world.block.Block;
-import org.terasology.engine.world.block.BlockComponent;
-import org.terasology.engine.world.block.BlockManager;
-import org.terasology.engine.world.block.BlockRegion;
-import org.terasology.engine.world.block.family.BlockPlacementData;
-import org.terasology.engine.world.block.regions.BlockRegionComponent;
-
-@RegisterSystem
-public class CardSystem extends BaseComponentSystem {
```
skaldarnar: Looking at this code, I'm not even sure anymore whether the cards act as doors or not :thinking:

skaldarnar (dismissed):

src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
skaldarnar: P.S. we just found out that our door placement itself is broken right now :eyes: :door: :question:

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/LASSystem.java
ahv15: The main reason I am removing this is because, this doesn't allow us to choose our starting inventory based on the defined prefab. All it does is remove all the items in the players inventory which doesn't give the player anything to fight with. By removing this we can settle on a more well defined starting inventory via the prefab. I will shift this to another PR. If this were to run in the end it would probably make sure that the player would start without anything in their inventory rather than what was defined to be the starting inventory.

src/main/java/org/terasology/ligthandshadow/logic/CardComponent.java
ahv15: Yup I noticed this but I assumed that the assets would be of some use in the future, therefore I didn't remove them. However there were some other components which I am not too sure serve any purpose like the TeleporterComponent for example and it seems like all it is doing is indicating that the dice acts as a teleporter.

src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
ahv15: So in that case would it be best to remove this for the time being and try and fix card placement to use as walls as a part of a later PR?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/LASSystem.java
jdrueckert: I think the intention behind this code was for the players to not have any items in their inventory in the lobby stage. Generally I do think that makes sense as the actual game takes place in the arena and not in the lobby.

An alternative for the lobby stage might be to make players invincible. Thus, players could already test out the items without being able to kill each other at that stage yet. However, for placeable (traps) / consumable (ammunition) items, we would need to ensure they still start into the game with the full set.

Shifting this to another PR makes sense, so we can discuss this there in more detail :+1:

src/main/java/org/terasology/ligthandshadow/logic/CardComponent.java
jdrueckert: To resolve the incorrect dependency "direction", maybe we should rather move this to LASR.

src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
jdrueckert: I'd vote for keeping it (potentially considering to move it to LASR too) and writing an issue tracking to replace this leveraging `MultiBlock` instead.

ahv15: In that case should I retain CardComponent too.

jdrueckert: yes, but move it to LASR to resolve the dependency being the wrong way round

jdrueckert (approved):

PR #212, opened 2021-06-12
Fix dependency issues due to components

ahv15:
Requires https://github.com/Terasology/LightAndShadowResources/pull/65 to test this out.
- [x] Moves components that are instantiated only by prefabs in LASResources to LASResources.


jdrueckert (commented):
You'll need to adjust the imports otherwise this will not compile even if the corresponding LASR PR is merged.

jdrueckert (approved):

PR #213, opened 2021-06-13
refactor: move flag logic to FlagAuthoritySystem

ahv15:
Currently a lot of the functionalities which handles flags is scattered around in separate files. This PR moves all those functions to FlagAuthoritySystem making it easier to categorize classes. Requires https://github.com/Terasology/LightAndShadowResources/pull/66 to test it


src/main/java/org/terasology/ligthandshadow/componentsystem/events/GiveFlagEvent.java
```diff
@@ -0,0 +1,15 @@
...
+import org.terasology.engine.entitySystem.event.Event;
+
+
+public class GiveFlagEvent implements Event {
+    public EntityRef flag;
+
+    public GiveFlagEvent(EntityRef flag) {
+        this.flag = flag;
+    }
+}
```
jdrueckert: newline

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -146,6 +103,17 @@ private boolean itemIsFlag(EntityRef checkedItem) {
...
+    private String checkWhichFlagPicked(InventorySlotChangedEvent event) {
+        item = event.getNewItem();
+        if (item.hasComponent(BlackFlagComponent.class)) {
+            return LASUtils.BLACK_TEAM;
+        }
+        if (item.hasComponent(RedFlagComponent.class)) {
+            return LASUtils.RED_TEAM;
+        }
+        return null;
+    }
```
jdrueckert: I wonder whether we might want to drop the `BlackFlag` and `RedFlag` components and rather have a component class member `team` or so, that holds the respective URI we return here...

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/FlagAuthoritySystem.java
```diff
@@ -113,4 +120,38 @@ public void onGiveItemToCharacterHoldItem(GiveItemEvent event, EntityRef item, I
                 delayManager.cancelDelayedAction(item, LASUtils.DROPPED_FLAG);
         }
     }
+
+    @ReceiveEvent
+    public void dropFlagRequest(FlagDropRequestEvent event, EntityRef targetPlayer) {
```
jdrueckert: naming nitpick: why "request" here and only here? theoretically, "move flag to base" and "give flag to player" also are mere requests. I'd say we drop the "Request" suffix here to keep it consistent.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/FlagAuthoritySystem.java
```diff
@@ -113,4 +120,38 @@ public void onGiveItemToCharacterHoldItem(GiveItemEvent event, EntityRef item, I
...
+        int inventorySize = inventoryManager.getNumSlots(targetPlayer);
+        for (int slotNumber = 0; slotNumber <= inventorySize; slotNumber++) {
+            EntityRef inventorySlot = inventoryManager.getItemInSlot(targetPlayer, slotNumber);
+            if (inventorySlot.hasComponent(BlockItemComponent.class)) {
+                if (inventorySlot.getComponent(BlockItemComponent.class).blockFamily.getURI().toString().equals(event.flagTeam)) {
+                    flagSlot = inventorySlot;
+                }
```
jdrueckert: you could consider working with optionals here to avoid first checking "has" and then doing "get". with an optional you could do "get" and "ifPresent" continue to get the block family, etc.

jdrueckert (dismissed):
IMO all the newly introduced events are trigger events.
Please align them with what's documented in https://github.com/MovingBlocks/Terasology/wiki/Event-Types#trigger-events, especially them being immutable.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/FlagAuthoritySystem.java
ahv15: The main reason I did that over here is because of the overlap with the FlagDrop event. The FlagDrop event is triggered when the flag drops and to avoid any conflicts I added the Request suffix to indicate that the flag should be dropped when this event is sent.

jdrueckert: The "request" to drop a flag would be a trigger event while the "flag drops now / was just dropped" is more of a notification event. Let's refer to the documentation linked above and adjust the naming and event implementation accordingly, I'd say.

jdrueckert (commented):
Please add documentation - especially for the newly introduced events and event handlers. You can have a look at the damage-unrelated parts of `Health` for some examples.

src/main/java/org/terasology/ligthandshadow/componentsystem/events/FlagDropEvent.java
```diff
@@ -1,32 +1,27 @@
...
+// SPDX-License-Identifier: Apache-2.0
 package org.terasology.ligthandshadow.componentsystem.events;
 
 import org.terasology.engine.entitySystem.entity.EntityRef;
 import org.terasology.engine.entitySystem.event.Event;
-import org.terasology.engine.network.BroadcastEvent;
 
-@BroadcastEvent
+/**
+ * This event causes the player to drop the flag.
```
jdrueckert: I think it would make sense to include the event type in the documentation.

src/main/java/org/terasology/ligthandshadow/componentsystem/events/FlagDropEvent.java
```diff
@@ -1,32 +1,27 @@
...
+ */
 public class FlagDropEvent implements Event {
```
jdrueckert: To give this a bit more of a "trigger" characteristic and align it with `GiveFlagEvent`, I'd propose to rename it to `DropFlagEvent`

src/main/java/org/terasology/ligthandshadow/componentsystem/events/OnFlagPickupEvent.java
```diff
@@ -20,14 +20,14 @@
 import org.terasology.engine.network.BroadcastEvent;
 
 @BroadcastEvent
-public class FlagPickupEvent implements Event {
+public class OnFlagPickupEvent implements Event {
     public String team;
     public EntityRef player;
```
jdrueckert: We could make these private and introduce getters :thinking:

jdrueckert (commented):
I didn't test this yet, but code-wise I only saw a few minor things.
@skaldarnar is there anything else you'd want to be addressed?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -77,33 +50,17 @@ private void dropFlagOnPlayerAttack(ActivateEvent event, EntityRef targetPlayer)
             if (targetPlayer.hasComponent(PlayerCharacterComponent.class) && targetPlayer.hasComponent(HasFlagComponent.class)) {
                 // If the target player has the black flag
                 if (targetPlayer.getComponent(HasFlagComponent.class).flag.equals(LASUtils.BLACK_TEAM)) {
-                    dropFlag(targetPlayer, attackingPlayer, LASUtils.BLACK_FLAG_URI);
+                    targetPlayer.send(new DropFlagEvent(attackingPlayer, LASUtils.BLACK_FLAG_URI));
```
skaldarnar: A few things to consider here (for a follow-up PR) I already discussed with @jdrueckert offline:

Don't send the specific item URI in the event, but only the team reference. This decouples the item/prefab handling from the game logic even more.

skaldarnar: What is the _intent_ of the `AttackSystem` here? And what are the _invariants_ that hold for our components? I think the attack system is only interested in stating that the player should drop the (enemy) flag. 

At the same time, a player can have either 0 or 1 flags because:
- an entity can have at most one instance of a component, i.e., there is either no or one `HasFlagComponent`, and
- `HasFlagComponent` has a single field for `flag`.

Given that, we should be able to resolve which flag to drop in the event handler for `DropFlagEvent` without the additional flag URI.

skaldarnar: These questions uncover a few assumptions or uncertainties with the current code that we should document. 

For instance, is the `HasFlagComponent#flag` always the enemy flag or could it be the flag of the own team?
Even if a player could pick up their own flag, would it be tracked in `HasFlagComponent`? More general, what is the _intent_ of `HasFlagComponent`?

skaldarnar: (_Disclaimer: We could not agree what the best way forward for the latter points is._)

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -127,12 +84,11 @@ public void onInventorySlotChanged(InventorySlotChangedEvent event, EntityRef en
         if (itemIsFlag(item)) {
             String flagTeam = checkWhichFlagPicked(event);
             if (flagTeam.equals(player.getComponent(LASTeamComponent.class).team)) {
-                moveFlagToBase(player, flagTeam);
-                return;
+                player.send(new MoveFlagToBaseEvent(item, flagTeam));
```
skaldarnar: Why do we send this event against the player? I would have assumed that

- we send the event against the flag item (because that's what we want to move)
- `ReturnToBaseEvent` (yes, different name) would return the target to its base
- the target base is implicitly given by the team the flag belongs to

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -127,12 +84,11 @@ public void onInventorySlotChanged(InventorySlotChangedEvent event, EntityRef en
...
             } else {
                 handleFlagPickup(player, flagTeam);
-                return;
             }
+            return;
```
skaldarnar: Not sure about my Java skills, but do we really need this for `void` methods? The `return` was previously only used to break early (unnecessarily).

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -143,11 +99,19 @@ public void onInventorySlotChanged(InventorySlotChangedEvent event, EntityRef en
     }
 
     private boolean itemIsFlag(EntityRef checkedItem) {
-        return (checkedItem.hasComponent(BlackFlagComponent.class) || checkedItem.hasComponent(RedFlagComponent.class));
+        return checkedItem.hasComponent(FlagComponent.class);
+    }
+
+    private String checkWhichFlagPicked(InventorySlotChangedEvent event) {
```
skaldarnar, line 99: This method should have Javadoc to state what values it can possibly return.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/FlagAuthoritySystem.java
```diff
@@ -113,4 +122,58 @@ public void onGiveItemToCharacterHoldItem(GiveItemEvent event, EntityRef item, I
...
+        Vector3f impulse = event.getAttackingPlayer().getComponent(LocationComponent.class).getLocalPosition()
+                .add(startPosition, new Vector3f()).div(2f);
+        targetPlayer.send(new DropItemRequest(flagSlot, targetPlayer, impulse, startPosition));
+    }
+
+    /**
+     * Places the flag at the base and removes the flag from the players inventory.
+     *
+     * @param event The MoveFlagToBase event sent to the player.
+     * @param playerEntity The entity from which the flag should be removed.
```
skaldarnar, line 156: Ah, now I get why this event is sent against the player and not the flag ... :thinking:

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -203,17 +190,10 @@ private void resetLevel() {
                 continue;
             }
 
-            movePlayerFlagToBase(playerWithFlag, oppositionTeam, heldFlag);
+            playerWithFlag.send(new MoveFlagToBaseEvent(heldFlag,oppositionTeam));
```
skaldarnar: Ah, we use the same event for "**return** flag to own base" and "**reset** flag on score"... In both cases, the flag is returned to the base of the team the flag represents, so I still think we don't need the target base as explicit parameter.

src/main/java/org/terasology/ligthandshadow/componentsystem/events/DropFlagEvent.java
```diff
@@ -0,0 +1,27 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.ligthandshadow.componentsystem.events;
+
+import org.terasology.engine.entitySystem.entity.EntityRef;
+import org.terasology.engine.entitySystem.event.Event;
+
+/**
+ * This trigger event causes the player to drop the flag.
```
skaldarnar, line 9: :+1: good to see the event type categorization here.

src/main/java/org/terasology/ligthandshadow/componentsystem/events/MoveFlagToBaseEvent.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.terasology.engine.entitySystem.entity.EntityRef;
+import org.terasology.engine.entitySystem.event.Event;
+
+/**
+ * This trigger event is responsible for sending the flag back to the base.
+ */
+public class MoveFlagToBaseEvent implements Event {
+    private final EntityRef heldFlag;
+    private final String flagTeam;
```
skaldarnar: I would call this event `ReturnFlagEvent` and omit the `flagTeam` property, as it is implicit that the flag should be returned to its home base.

src/main/java/org/terasology/ligthandshadow/componentsystem/events/MoveFlagToBaseEvent.java
```diff
@@ -0,0 +1,27 @@
...
+package org.terasology.ligthandshadow.componentsystem.events;
+
+import org.terasology.engine.entitySystem.entity.EntityRef;
+import org.terasology.engine.entitySystem.event.Event;
+
+/**
+ * This trigger event is responsible for sending the flag back to the base.
+ */
+public class MoveFlagToBaseEvent implements Event {
+    private final EntityRef heldFlag;
```
skaldarnar: Why did you chose the name `heldFlag` here?

skaldarnar (commented):
I think this is an improvement of the code, thus we can continue with merging it :+1: 

However, I've left a bunch of comments we should go over to see which of them we want to address (when).

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
ahv15: HasFlagComponent should mainly be used to indicate that the player has the opponents flag and indicating which team the flag the player has is from is unnecessary as the flagComponent already does that.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -75,35 +47,12 @@ private void dropFlagOnPlayerAttack(ActivateEvent event, EntityRef targetPlayer)
...
-                if (targetPlayer.getComponent(HasFlagComponent.class).flag.equals(LASUtils.BLACK_TEAM)) {
-                    dropFlag(targetPlayer, attackingPlayer, LASUtils.BLACK_FLAG_URI);
-                    return;
-                }
-                if (targetPlayer.getComponent(HasFlagComponent.class).flag.equals(LASUtils.RED_TEAM)) {
-                    dropFlag(targetPlayer, attackingPlayer, LASUtils.RED_FLAG_URI);
-                    return;
-                }
+                // If the target player has the flag
+                targetPlayer.send(new DropFlagEvent(attackingPlayer));
```
jdrueckert, line 51: why do we send the attacking player here?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/FlagAuthoritySystem.java
```diff
@@ -113,4 +123,59 @@ public void onGiveItemToCharacterHoldItem(GiveItemEvent event, EntityRef item, I
...
+            Optional<FlagComponent> flagComponent = Optional.ofNullable(inventoryManager.getItemInSlot(targetPlayer,
+                    slotNumber).getComponent(FlagComponent.class));
+            flagComponent.ifPresent(flag -> {
+                if (flag.team.equals(LASUtils.getOppositionTeam(targetPlayer.getComponent(LASTeamComponent.class).team))) {
+                    flagSlot = inventorySlot;
+                }
+            });
+        }
+        Vector3fc startPosition = targetPlayer.getComponent(LocationComponent.class).getLocalPosition();
+        Vector3f impulse = event.getAttackingPlayer().getComponent(LocationComponent.class).getLocalPosition()
```
jdrueckert, line 147: ah, I see we need the attacking player to calculate the impulse... seems a bit weird but oh well...

jdrueckert (approved):

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
skaldarnar: I guess we could decide to compute the direction from which the magic staff "activation" comes here in the attack system and only pass that to the `DropFlagEvent`. 

On the other hand, this might be a nice addition to the score board if you think for "scorer points" from other sports, e.g., giving players credit for defending their flag...

PR #214, opened 2021-06-15
fix: non-deterministic inventory and empty respawn inventory

### LINKED ISSUE #109: After being killed by another player you respawn with an completely empty inventory (no weapons)
In the Teracon 2019 multiplayer playtest we encountered issues with the combat system. 

- Empty inventory and toolbar (Tuesday 15. Oct)
When using the latest master of before the teracon and applying the PR #107 we encountered the issue, that when you are being killed by another player you respawn with an empty inventory and an empty toolbar so you cannot fight anymore. 

- An additional quirk i cannot remember correctly right now, highlighting @skaldarnar in the hope that he can add his part for that one.

### LINKED ISSUE #117: Nondeterministic Starting Inventory
Sometimes the player starts with a nearly empty inventory with only the LaS Staff, and sometimes the player is spawned with the starting inventory of Combat System (various weapons).

ahv15:
Requires https://github.com/Terasology/Inventory/pull/44 to test this PR.  This PR adds items to the players inventory when the player spawns at the base or respawns on death. It makes use of the default StartingInventory used in CombatSystem for inventory but this can be modified and finalized once the weapons are integrated with effects later on.
Fixes https://github.com/Terasology/LightAndShadow/issues/117
Fixes https://github.com/Terasology/LightAndShadow/issues/109


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -83,6 +87,10 @@ public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterC
             player.send(new RestoreFullHealthEvent(player));
             Vector3f randomVector = new Vector3f(-1 + random.nextInt(3), 0, -1 + random.nextInt(3));
             player.send(new CharacterTeleportEvent(randomVector.add(LASUtils.getTeleportDestination(team))));
+            Optional<Prefab> prefab = Assets.getPrefab("inventory");
+            StartingInventoryComponent startingInventory = prefab.get().getComponent(StartingInventoryComponent.class);
+            player.addOrSaveComponent(startingInventory);
+            player.send(new RequestInventoryEvent());
```
skaldarnar: Instead of adding the component and then triggering the action with the empty event I'd suggest to add the payload to the event itself, see https://github.com/Terasology/Inventory/pull/44#pullrequestreview-684342609.

This would just become 
```suggestion
            player.send(new RequestInventoryEvent(startingInventory.items));
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -83,6 +87,10 @@ public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterC
             player.send(new RestoreFullHealthEvent(player));
             Vector3f randomVector = new Vector3f(-1 + random.nextInt(3), 0, -1 + random.nextInt(3));
             player.send(new CharacterTeleportEvent(randomVector.add(LASUtils.getTeleportDestination(team))));
+            Optional<Prefab> prefab = Assets.getPrefab("inventory");
+            StartingInventoryComponent startingInventory = prefab.get().getComponent(StartingInventoryComponent.class);
```
skaldarnar: We resolve and load the asset every time a player dies. As the respawn inventory does not change, you could just initialize this once for the system and then reuse the value every time you reset the inventory.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -116,6 +121,10 @@ private String setPlayerTeamToTeleporterTeam(EntityRef player, EntityRef telepor
     private void handlePlayerTeleport(EntityRef player, String team) {
         Vector3f randomVector = new Vector3f(-1 + random.nextInt(3), 0, -1 + random.nextInt(3));
         player.send(new CharacterTeleportEvent(randomVector.add(LASUtils.getTeleportDestination(team))));
+        Optional<Prefab> prefab = Assets.getPrefab("inventory");
+        StartingInventoryComponent startingInventory = prefab.get().getComponent(StartingInventoryComponent.class);
+        player.addComponent(startingInventory);
+        player.send(new RequestInventoryEvent());
```
skaldarnar: Having the same code twice indicates that you probably should refactor this. You might want to consider introducing a LaS-specific event denoting a "player reset", and then fire that event both when the player teleports into the game and when they respawn.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -116,6 +121,10 @@ private String setPlayerTeamToTeleporterTeam(EntityRef player, EntityRef telepor
...
         inventoryManager.giveItem(player, EntityRef.NULL, entityManager.create(LASUtils.MAGIC_STAFF_URI));
```
skaldarnar: Why is the magic staff added separately and not part of the respawn inventory?

ahv15: The magic staff isn't dropped on death so the player get's it back on respawn. Over here we are using the same inventory for spawn and respawn and that would mean that on respawn the player will have multiple staffs, I could create separate inventories for spawn and respawn but most of the weapons are similar currently and changes to the inventory will be coming later on.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: As we will be initializing the startingInventory in the beginning of the systems and the inventories should eventually be different for spawn and respawn that would leave only the last 2 lines to be refactored which might be unnecessary.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -94,7 +103,9 @@ private void dropItemsFromInventory(EntityRef player) {
         for (int slotNumber = 0; slotNumber <= inventorySize; slotNumber++) {
             EntityRef slot = inventoryManager.getItemInSlot(player, slotNumber);
             Prefab currentPrefab = slot.getParentPrefab();
-            if (currentPrefab != null && !currentPrefab.equals(staffPrefab)) {
+            if (currentPrefab != null && currentPrefab.equals(staffPrefab)) {
+                inventoryManager.removeItem(player, EntityRef.NULL, slotNumber, true, 1);
```
jdrueckert: I don't think we should treat the staff different than every other item in the inventory here.
I'm aware that there is a task where we will need to discuss which items to treat how (keep / drop / remove from the game), but I think when we do that we should do it for all items and not start doing it for the staff already now.
Hence, I would drop the staff on death like all the other items.

assets/prefabs/inventory.prefab
```diff
@@ -0,0 +1,17 @@
...
+      { "uri": "CombatSystem:sword", "quantity": 1 },
+      { "uri": "CombatSystem:waraxe", "quantity": 1 },
+      { "uri": "CombatSystem:staff", "quantity": 1 },
+      { "uri": "CombatSystem:spearItem", "quantity": 1 },
+      { "uri": "CombatSystem:fireBallMine", "quantity": 1 },
+      { "uri": "CombatSystem:explodeMine", "quantity": 1 },
+      { "uri": "LightAndShadowResources:magicStaff", "quantity": 1 }
+    ]
+  }
+}
```
jdrueckert: newline please

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
ahv15: I didn't want to remove the logic implemented previously, wherein the staff is not dropped on death, but now as we are giving the staff too on respawn I removed it every time the player dies. I could remove this in case the previously implemented logic is not what is required.

jdrueckert: I think until we think it through properly, which items should be kept, dropped, or removed, we should not have additional logic treating some different than others. For me this includes the logic that already was there. Thus, at least we have consistent behaviour for all items.

jdrueckert (approved):
Looks fine from my pov.

PR #216, opened 2021-06-20
fix: update surfaces facet to prevent tree's in certain locations

ahv15:
Currently tree's might spawn in certain locations preventing the generation of certain parts of the world. By updating the surfacesFacet the tree's should no longer spawn in those regions. 

Use the following seeds
- `1YWCPv3iajQCSM5W31U5cpNHE0Qdm4b3` to test whether it works for the bases, 
- `4BUJ2seIUjKOOORRgB90RtBrFGz6czBd` to test whether it works for the yin yang platform, 
- `NiVVGqjYgBDsaLanlH0f15wvaFGJb5QW` for the floating platform.

Closes #202


src/main/java/org/terasology/las/bases/BaseProvider.java
```diff
@@ -43,12 +48,22 @@ public void setSeed(long seed) {
...
 
         for (Base base : fixedBases) {
             facet.add(base);
         }
+
+        BlockRegion surfaceRegion = surfacesFacet.getWorldRegion();
+        for (Vector3ic pos : surfaceRegion) {
+            if ((redBaseRegion.contains(pos) || blackBaseRegion.contains(pos))) {
+                int y = surfacesFacet.getNextBelow(pos);
+                surfacesFacet.setWorld(pos.x(), y, pos.z(), false);
```
skaldarnar, line 64: I'm not sure what would happen if the `y` value is not part of the generating `region`. The implementation looks like it would not crash, but this might be a pitfall if the next surface layer is below the currently generated chunk...

src/main/java/org/terasology/las/yinyang/YinYangProvider.java
```diff
@@ -5,35 +5,56 @@
...
         yinYangPositions.stream()
                 .filter(worldRect::contains)
-                .forEach(pos -> yinYangFacet.setWorld(pos, new YinYang()));
+                .forEach(pos -> {
+                    yinYangFacet.setWorld(pos, new YinYang());
+                    int y = surfacesFacet.getNextBelow(pos);
+                    for (int i = -RADIUS; i <= RADIUS; i++) {
+                        for (int j = -2 * RADIUS; j <= 2 * RADIUS; j++) {
+                            if (!LASUtils.pixel(j, i, RADIUS).equals("engine:air")) {
+                                Vector3i blockPos = new Vector3i(i, 0, j).add(pos);
```
skaldarnar: The yin yang symbol is not that big, so the impact of creating new vectors here might not be that big, but it's still good practice to create a new instance once outside the loop (e.g., before `yinYangPositions.stream()`) and then `set` this temporary vector here.

We could also directly write the result of the addition to that temporary vector, like so:
```suggestion
                                pos.add(i, 0, j, tempPos);
```

src/main/java/org/terasology/las/yinyang/YinYangProvider.java
```diff
@@ -5,35 +5,56 @@
...
+                                blockPos.y = y;
+                                surfacesFacet.setWorld(blockPos, false);
```
skaldarnar: Again, I'm not sure what happens if you try to set the facet value for a position that is not contained in the generating `region`. Similar to the pitfall mentioned in the previous comment, this time leaving the region to the sides.

src/main/java/org/terasology/las/bases/BaseProvider.java
```diff
@@ -43,12 +48,22 @@ public void setSeed(long seed) {
...
+            }
+        }
         region.setRegionFacet(BaseFacet.class, facet);
+        region.setRegionFacet(SurfacesFacet.class, surfacesFacet);
```
naalit: This is unnecessary: there's only one copy of the facet, which was modified in-place. This is only needed if the provider is creating a new facet.

src/main/java/org/terasology/las/bases/BaseProvider.java
```diff
@@ -43,12 +48,22 @@ public void setSeed(long seed) {
...
+        SurfacesFacet surfacesFacet = region.getRegionFacet(SurfacesFacet.class);
         Border3D border = region.getBorderForFacet(BaseFacet.class);
         BaseFacet facet = new BaseFacet(region.getRegion(), border);
 
         for (Base base : fixedBases) {
             facet.add(base);
         }
+
+        BlockRegion surfaceRegion = surfacesFacet.getWorldRegion();
+        for (Vector3ic pos : surfaceRegion) {
```
naalit: The base regions are a lot smaller, so I think it would make more sense to loop over those and then check if it's in the generating region. Maybe:

```suggestion
        for (Vector3ic pos : Iterables.concat(redBaseRegion, blackBaseRegion)) {
            if (surfaceRegion.contains(pos)) {
```

src/main/java/org/terasology/las/yinyang/YinYangProvider.java
```diff
@@ -5,35 +5,56 @@
...
+                            }
+                        }
+                    }
+                });
 
         region.setRegionFacet(YinYangFacet.class, yinYangFacet);
+        region.setRegionFacet(SurfacesFacet.class, surfacesFacet);
```
naalit: As above, this is unnecessary.

```suggestion
```

src/main/java/org/terasology/las/yinyang/YinYangProvider.java
ahv15: https://www.baeldung.com/java-8-lambda-expressions-tips#mutation

src/main/java/org/terasology/las/platform/FloatingPlatformProvider.java
```diff
@@ -77,7 +83,15 @@ public void process(GeneratingRegion region) {
...
-
+        BlockRegion worldRegion = surfacesFacet.getWorldRegion();
+        for (int x = worldRegion.minX(); x <= worldRegion.maxX(); x++) {
+            for (int z = worldRegion.minZ(); z <= worldRegion.maxZ(); z++) {
+                int y = surfacesFacet.getNextBelow(new Vector3i(x, LASUtils.FLOATING_PLATFORM_HEIGHT_LEVEL, z));
+                if (worldRegion.contains(x, y, z) && FLOATING_PLATFORM_REGION.contains(x, z)) {
+                    surfacesFacet.setWorld(x, y, z, false);
+                }
+            }
+        }
```
skaldarnar: This creates a new `Vector3i` each time, which is unnecessarily expensive. It would already be a lot better if we'd reuse the same temporary vector.

```suggestion
        Vector3i pos = new Vector3i();
        for (int x = worldRegion.minX(); x <= worldRegion.maxX(); x++) {
            for (int z = worldRegion.minZ(); z <= worldRegion.maxZ(); z++) {
                pos.set(x, LASUtils.FLOATING_PLATFORM_HEIGHT_LEVEL, z);
                int y = surfacesFacet.getNextBelow(pos);
                if (worldRegion.contains(x, y, z) && FLOATING_PLATFORM_REGION.contains(x, z)) {
                    surfacesFacet.setWorld(x, y, z, false);
                }
            }
        }
```

However, it would be even better to simply add the respective variants taking three integer arguments to the `SurfacesFacet`. The values are "unpacked" in those methods anyway, so I think this is reasonable addition.

```suggestion
        for (int x = worldRegion.minX(); x <= worldRegion.maxX(); x++) {
            for (int z = worldRegion.minZ(); z <= worldRegion.maxZ(); z++) {
                int y = surfacesFacet.getNextBelow(x, LASUtils.FLOATING_PLATFORM_HEIGHT_LEVEL, z);
                if (worldRegion.contains(x, y, z) && FLOATING_PLATFORM_REGION.contains(x, z)) {
                    surfacesFacet.setWorld(x, y, z, false);
                }
            }
        }
```

src/main/java/org/terasology/las/yinyang/YinYangProvider.java
```diff
@@ -5,33 +5,54 @@
...
                         .extendBy(0, 0, 10);
 
+        SurfacesFacet surfacesFacet = region.getRegionFacet(SurfacesFacet.class);
+
         YinYangFacet yinYangFacet = new YinYangFacet(region.getRegion(), border);
 
         BlockRegion worldRect = yinYangFacet.getWorldRegion();
-
+        Vector3i[] tempPos = new Vector3i[1];
+        tempPos[0] = new Vector3i();
```
skaldarnar: Why do you use an array with a single element here? You could just use the temp vector directly :thinking:

skaldarnar (changes requested):

src/main/java/org/terasology/las/yinyang/YinYangProvider.java
ahv15: https://www.baeldung.com/java-8-lambda-expressions-tips#mutation

ahv15: That was required to make use of mutable variables in a lambda expression.

skaldarnar: That's not exactly a "best practices" example, but rather a showcase how the "effectively final" rule  is helpful, but not sufficient to prevent accidental errors. 

That said, we luckily have a better option here: Using a similar trick, we'll use a final `Vector3i` (we won't change the reference/object pointer) and only `.set(...)` its value in the lambda expression.

Note that we never do a reassignment `tmp = ...` but only mutation through method calls `tmp.set(...)`. This also helps to keep the number of instantiations low.

src/main/java/org/terasology/las/platform/FloatingPlatformProvider.java
ahv15: looks like it doesn't get unpacked as the only argument to that function is a vector, did you mean adding a separate function which uses the separated values directly?

skaldarnar (approved):

PR #218, opened 2021-06-23
refactor: remove unused code

ahv15:
Removes unused components, SetTeamAction and the UI part from an Authority System (which has already been taken care by a Client System). Requires https://github.com/Terasology/LightAndShadowResources/pull/67.


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -38,50 +23,18 @@
...
-                return String.valueOf(redScore);
-            }
-        });
-    }
-
-    @Override
-    public void initialise() {
-        // Displays score UI on game start
-        nuiManager.getHUD().addHUDElement("ScoreHud");
-    }
```
jdrueckert, line 83: Did you miss adding the file you moved this to to the commit? 🤔

jdrueckert (changes requested):
The particle components should not be removed as they are used by the particle effect prefabs in https://github.com/Terasology/LightAndShadowResources/tree/develop/assets/prefabs/cards

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
ahv15: This is already a part of ClientScoreSystem.
https://github.com/Terasology/LightAndShadow/blob/33c72bc232963c110324563b61c922fec07d2137/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientScoreSystem.java#L43-L70

ahv15: Although those components are a part of those prefabs, no system is utilizing them so they are not really required.  We should remove these components from those prefabs too.

jdrueckert: > Although those components are a part of those prefabs, no system is utilizing them so they are not really required. We should remove these components from those prefabs too.

Have you compared carrying a flag in-game with and without the PR?

ahv15: Yup I tested everything and it worked fine, while picking up the flags the particles were rendered. Tested it in multiplayer too and it works fine.

jdrueckert (commented):
Indeed everything seems to work as expected in-game...
Did you already create the associated LASR PR that removes the removed components from prefabs?

ahv15:
Not yet just wanted to confirm if we could move ahead with this.

jdrueckert (approved):

PR #219, opened 2021-06-26
feat: add pregame phase

ahv15:
Creates a pregame phase which gives players time to join teams and setup their defenses. Requires https://github.com/Terasology/LightAndShadowResources/pull/68 . For the actual game to start there has to be at least one player in each team.


src/main/java/org/terasology/ligthandshadow/componentsystem/components/PlayerInvulnerableComponent.java
```diff
@@ -0,0 +1,8 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.ligthandshadow.componentsystem.components;
+
+import org.terasology.engine.entitySystem.Component;
+
+public class PlayerInvulnerableComponent implements Component {
```
jdrueckert: I'd probably go for a `InvulnerableComponent` here as we potentially could use that in the future also for NPCs and stuff.

deltas/CombatSystem/prefabs/projectiles/basicShoot.prefab
```diff
@@ -0,0 +1,5 @@
+{
+    "LASTeam" : {
+        "team" : "white"
+    }
+}
```
jdrueckert: Why do we need this?

src/main/java/org/terasology/ligthandshadow/componentsystem/events/BarrierActivateEvent.java
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.ligthandshadow.componentsystem.events;
+
+import org.terasology.engine.entitySystem.event.Event;
+
+public class BarrierActivateEvent implements Event {
```
jdrueckert: Are these (activate and deactivate) notification or trigger events? You might want to indicate that with the naming (and documentation): `ActivateBarrierEvent` vs `OnBarrierActivateEvent`

src/main/java/org/terasology/ligthandshadow/componentsystem/events/BarrierDeactivateEvent.java
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.ligthandshadow.componentsystem.events;
+
+import org.terasology.engine.entitySystem.event.Event;
+
+public class BarrierDeactivateEvent implements Event {
```
jdrueckert: see above

src/main/java/org/terasology/ligthandshadow/componentsystem/events/GameStartMessageEvent.java
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.ligthandshadow.componentsystem.events;
+
+import org.terasology.engine.entitySystem.event.Event;
+import org.terasology.engine.network.OwnerEvent;
+
+@OwnerEvent
+public class GameStartMessageEvent implements Event {
```
jdrueckert, line 12: without having looked into the systems that handle this event yet, it's unclear to me what this is used for

src/main/java/org/terasology/ligthandshadow/componentsystem/events/PregameEvent.java
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.ligthandshadow.componentsystem.events;
+
+import org.terasology.engine.entitySystem.event.Event;
+
+public class PregameEvent implements Event {
```
jdrueckert, line 10: see above

src/main/java/org/terasology/ligthandshadow/componentsystem/events/TimerEvent.java
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+package org.terasology.ligthandshadow.componentsystem.events;
+
+import org.terasology.engine.entitySystem.event.Event;
+import org.terasology.engine.network.OwnerEvent;
+
+@OwnerEvent
+public class TimerEvent implements Event {
```
jdrueckert, line 12: see above

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
+import org.terasology.ligthandshadow.componentsystem.LASUtils;
+import org.terasology.ligthandshadow.componentsystem.events.BarrierActivateEvent;
+import org.terasology.ligthandshadow.componentsystem.events.BarrierDeactivateEvent;
+import org.terasology.module.inventory.systems.InventoryManager;
 
-@RegisterSystem
+@RegisterSystem(RegisterMode.AUTHORITY)
 public class MagicDomeSystem extends BaseComponentSystem implements UpdateSubscriberSystem {
-
+    private static final int WORLD_RADIUS = 20;
```
jdrueckert: this is no longer a "world" radius but rather a "pregame zone" radius

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
-    public void postBegin() {
-        //toggleDome();
-    }
 
-    @Command(shortDescription = "Activate/Deactivate dome barrier", helpText = "Activates or deactivates the dome barrier around the world", runOnServer = true)
-    public String dome() {
-        toggleDome();
-        return "Toggled dome.";
+    @ReceiveEvent
+    public void activateBarrier(BarrierActivateEvent event, EntityRef entity) {
```
jdrueckert:
```suggestion
    public void activateBarriers(BarrierActivateEvent event, EntityRef entity) {
```

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
-            AnimateRotationComponent rotationComponent = new AnimateRotationComponent();
-            rotationComponent.rollSpeed = 0.01f;
-            rotationComponent.pitchSpeed = 0.01f;
-            rotationComponent.yawSpeed = 0.01f;
-            magicDomeEntity.addOrSaveComponent(rotationComponent);
-        } else {
-            magicDomeEntity.destroy();
-        }
+    @ReceiveEvent
+    public void deactivateBarrier(BarrierDeactivateEvent event, EntityRef entity) {
```
jdrueckert:
```suggestion
    public void deactivateBarriers(BarrierDeactivateEvent event, EntityRef entity) {
```

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
-
         for (EntityRef domeEntity : entityManager.getEntitiesWith(MagicDome.class, LocationComponent.class)) {
             LocationComponent domeLocationComponent = domeEntity.getComponent(LocationComponent.class);
             Vector3f domeCenter = domeLocationComponent.getWorldPosition(new Vector3f());
-            Vector3f domeVerticalTop = domeCenter.add(0, WORLD_RADIUS, 0);
-
-            MagicDome dome = domeEntity.getComponent(MagicDome.class);
+            pos.sub(domeCenter, df);
+            float distance = df.length();
+//          logger.info("CharacerMoveInputEvent: position: {} - distance from O: {}, delta: {}", pos, distance, deltaDistance);
```
jdrueckert: please remove these

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
     @In
     private EntityManager entityManager;
+    @In
+    private InventoryManager inventoryManager;
 
-    private Vector3f lastPos = new Vector3f();
-    private EntityRef magicDomeEntity = EntityRef.NULL;
     private float updateDelta;
-    private FastRandom random = new FastRandom();
+    private Vector3f df = new Vector3f();
```
jdrueckert: df?

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
+        if (updateDelta > 0.1f) {
+            // required to destroy the fireball sent by the staff
+            Iterable<EntityRef> fireballs = entityManager.getEntitiesWith(ExplodeComponent.class, LASTeamComponent.class);
+            destroyProjectile(fireballs);
+            // required to destroy the arrows and the spear
+            Iterable<EntityRef> arrows = entityManager.getEntitiesWith(ArrowComponent.class, LASTeamComponent.class);
+            destroyProjectile(arrows);
             updateDelta = 0;
         }
     }
```
jdrueckert: guess we don't need this anymore, do we?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -100,6 +101,10 @@ private boolean isProperTeamSize(EntityRef teleporter, EntityRef player) {
             }
         }
         if (teleporterTeamCount - oppositeTeamCount < maxTeamSizeDifference) {
+            if (teleporterTeamCount >= 0 && oppositeTeamCount >= 1 && !gameStart) {
+                sendTimerEventToClients();
+                gameStart = true;
+            }
```
jdrueckert: Not sure if this should be in `isProperTeamSize` :thinking: 
This seems to be a more or less unnecessary side effect.
Like not that the logic is unnecessary at all, but rather that it being a side effect here is unnecessary...

jdrueckert (changes requested):

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientPregameSystem.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public void shutdown() {
+        nuiManager.closeScreen(window);
+    }
+
+    @ReceiveEvent
+    public void onPregameStart(GameStartMessageEvent event, EntityRef entity) {
+        if (localPlayer.getClientEntity().equals(entity)) {
+            window.addNotification("The game start's when there is at least one player in each team.");
```
jdrueckert:
```suggestion
            window.addNotification("The game starts as soon as there is at least one player in each team.");
```

jdrueckert: this should be a constant

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientPregameSystem.java
```diff
@@ -0,0 +1,82 @@
...
+                        window.removeNotification("The game starts in " + (timePeriod + 1) + " seconds.");
+                        addNotification = true;
+                    }
+                    if (timePeriod < 0) {
+                        window.removeNotification("The game starts in " + (timePeriod + 1) + " seconds.");
+                        entity.send(new BarrierDeactivateEvent());
+                        entity.send(new RemoveInvulnerabilityEvent());
+                        timer.cancel();
+                    }
+                }
```
jdrueckert: let's have another look at this in a follow-up...

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
skaldarnar: This definitely needs some Javadoc.

- it activates the barriers for both teams
- but only if **both** barrier entities don't exist
- it will complete without any effect or log trace if any of the barriers is not the null entity

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Wouldn't iterating through the players again to get the count of players be code duplication.

jdrueckert: And suddenly, keeping this information as state seems more appealing again :see_no_evil: 
@skaldarnar what's your pov on this?

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
+        return barrier;
+    }
 
-                if (lastPos.length() > WORLD_RADIUS && lastPos.length() > pos.length() && distance < WORLD_RADIUS) {
-//                    logger.info("Sending player back outside");
-                    Vector3f impulse = new Vector3f(pos).normalize();
-                    float verticalDiff = Math.sqrt(Math.abs(Math.sqrt(lastPos.y()) - Math.sqrt(domeVerticalTop.y())));
-                    float impulseY = (Math.abs(verticalDiff) / (float) WORLD_RADIUS) * 3f;
+    public boolean isAllowedInsideBarrier(EntityRef player, EntityRef barrier) {
+        return barrier.getComponent(MagicDome.class).team.equals(player.getComponent(LASTeamComponent.class).team);
```
skaldarnar: This is a nice one-liner, but also quite error prone to show a NPE some day. To simplify debugging, you may consider splitting this up in multiple lines:

```suggestion
        String barrierTeam = barrier.getComponent(MagicDome.class).team;
        String playerTeam = player.getComponent(LASTeamComponent.class).team;
        return barrierTeam.equals(playerTeam);
```

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -28,119 +18,135 @@
...
-
-                entity.saveComponent(rotationComponent);
-            }
+        if (updateDelta > 0.1f) {
+            // required to destroy the fireball sent by the staff
+            Iterable<EntityRef> fireballs = entityManager.getEntitiesWith(ExplodeComponent.class, LASTeamComponent.class);
+            destroyProjectile(fireballs);
+            // required to destroy the arrows and the spear
+            Iterable<EntityRef> arrows = entityManager.getEntitiesWith(ArrowComponent.class, LASTeamComponent.class);
+            destroyProjectile(arrows);
```
skaldarnar: It would be nice to have a `ProjectileComponent` marker component here, I guess...I don't think we have any real concept for hierarchies like that yet.

src/main/java/org/terasology/ligthandshadow/componentsystem/components/PlayerInvulnerableComponent.java
skaldarnar: :gb: what's the difference between "invincible" and "invulnerable"? :us:  :thinking:

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientGameOverSystem.java
```diff
@@ -89,13 +80,13 @@ public void run() {
                         countDown.setText(Integer.toString(timePeriod--));
                         if (timePeriod < 0) {
                             countDown.setText(" ");
+                            restartButton.setEnabled(true);
```
skaldarnar, line 83: This just works by using the `timer`, no need for a custom `Thread`? Nice :+1:

src/main/java/org/terasology/ligthandshadow/componentsystem/events/BarrierActivateEvent.java
skaldarnar: Second one should use past tense: `OnBarrierActivatedEvent`

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -41,16 +45,13 @@
...
-    private static final Logger logger = LoggerFactory.getLogger(TeleporterSystem.class);
-
-    @In
-    InventoryManager inventoryManager;
     @In
     EntityManager entityManager;
     @In
     GameEntitySystem gameEntitySystem;
 
+    private boolean gameStart;
```
skaldarnar, line 52: What does this indicate?

skaldarnar: And is this value ever reset to `false` at the end of a game/when restarting?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -123,8 +128,28 @@ private String setPlayerTeamToTeleporterTeam(EntityRef player, EntityRef telepor
...
         Vector3f randomVector = new Vector3f(-1 + random.nextInt(3), 0, -1 + random.nextInt(3));
+        player.send(new PregameEvent());
         player.send(new CharacterTeleportEvent(randomVector.add(LASUtils.getTeleportDestination(team))));
         player.addOrSaveComponent(startingInventory);
         player.send(new RequestInventoryEvent(startingInventory.items));
+        sendGameStartMessageEventToClients();
+    }
+
+    private void sendTimerEventToClients() {
+        if (entityManager.getCountOfEntitiesWith(ClientComponent.class) != 0) {
```
skaldarnar: Why do you need this `if` check? If there are no clients, the iterable should be empty, and the `for` loop should effectively be a no-op.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -123,8 +128,28 @@ private String setPlayerTeamToTeleporterTeam(EntityRef player, EntityRef telepor
...
+            Iterable<EntityRef> clients = entityManager.getEntitiesWith(ClientComponent.class);
+            for (EntityRef client : clients) {
+                client.send(new TimerEvent());
+            }
+        }
+    }
+
+    private void sendGameStartMessageEventToClients() {
+        if (entityManager.getCountOfEntitiesWith(ClientComponent.class) != 0) {
```
skaldarnar: See above.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -123,8 +128,28 @@ private String setPlayerTeamToTeleporterTeam(EntityRef player, EntityRef telepor
...
+    private void sendTimerEventToClients() {
+        if (entityManager.getCountOfEntitiesWith(ClientComponent.class) != 0) {
+            Iterable<EntityRef> clients = entityManager.getEntitiesWith(ClientComponent.class);
+            for (EntityRef client : clients) {
+                client.send(new TimerEvent());
+            }
+        }
+    }
+
+    private void sendGameStartMessageEventToClients() {
```
skaldarnar: These `send...ToClients` helper functions have the same structure and purpose, the only difference is which event they are sending. 

You could refactor this to add the event [supplier](https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html) as an argument here:

```java
void sendEventToClients(Supplier<Event> eventSupplier) {
        Iterable<EntityRef> clients = entityManager.getEntitiesWith(ClientComponent.class);
        for (EntityRef client : clients) {
            client.send(eventSupplier.get());
        }
}
```

On caller side this would look like this:

```java
sendEventToClients(GameStartMessageEvent::new);
// ... which is syntactic sugar for 
sendEventToClients(() -> new GameStartMessageEvent());
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: This is just a check to make sure that the game hasn't started yet, if a player joins once the game starts the pregame timer would appear on their screen in case this check isn't there. This is only when the players teleport to the base, I haven't considered the pregame phase in case of restarts yet, but this check would only be needed if the restarted game doesn't have the same amount of players.

src/main/java/org/terasology/ligthandshadow/componentsystem/components/PlayerInvulnerableComponent.java
jdrueckert: If you're invulnerable you cannot be hurt/wounded, if you're invincible you cannot be defeated/killed.

src/main/java/org/terasology/ligthandshadow/componentsystem/events/GameStartMessageEvent.java
jdrueckert: if this is a notification event indicating that the pregame phase has started, why do we have `PregameEvent` that also indicates that the pregame phase has started :thinking:

ahv15: If I remember correctly this is run on the owner of the player whereas that is run on the player itself, we need to send the event to the owner so that the UI can be correctly configured and without this it doesn't work.

jdrueckert: ah okay, yeah that's possible of course.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
jdrueckert: So Skal just argued that looping through all players again would not be too much of a concern as we have - in comparison to the total number of entities - comparably few players.
@ahv15 can you maybe put a comment on the overhaul trello card and note to have another look at this as a follow-up?

jdrueckert (approved):

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
ahv15: Ok

PR #224, opened 2021-07-07
refactor: split restart system

ahv15:


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientRestartSystem.java
```diff
@@ -0,0 +1,42 @@
...
+    LocalPlayer localPlayer;
+    @In
+    NUIManager nuiManager;
+
+    @ReceiveEvent
+    public void onClientRestart(ClientRestartEvent event, EntityRef clientEntity) {
+        if (localPlayer.getClientEntity().equals(clientEntity)) {
+            if (nuiManager.isOpen(LASUtils.DEATH_SCREEN)) {
+                DeathScreen deathScreen = (DeathScreen) nuiManager.getScreen(LASUtils.DEATH_SCREEN);
+                MigLayout migLayout = deathScreen.find("playerStatistics", MigLayout.class);
```
skaldarnar: I don't think this needs to be that specific here, so let's improve it while touching the code. The method we need is defined in `UILayout`, so we can use that type instead of requiring it to be a `MigLayout`.

https://github.com/MovingBlocks/TeraNUI/blob/b7503ba0d733cc409df9bedf7fa284f8656f5b4d/nui/src/main/java/org/terasology/nui/UILayout.java#L41

skaldarnar (changes requested):

skaldarnar (approved):

PR #225, opened 2021-07-08
feat: add in-game statistics screen

ahv15:
Ingame statistics screen with notification to help player's use the screen for the first time. Requires a different icon for the notification.


assets/ui/statisticsScreen.ui
```diff
@@ -0,0 +1,289 @@
...
+                      },
+                      {
+                        "type": "UISpace"
+                      },
+                      {
+                        "type": "UISpace"
+                      },
+                      {
+                        "type": "UISpace"
+                      },
```
jdrueckert, line 202: Why exactly do we need to do it like this?

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientStatisticsSystem.java
```diff
@@ -0,0 +1,111 @@
...
+                })
+                .map(key -> FontColor.getColored(key, Color.yellow))
+                .orElse(FontColor.getColored("n/a", Color.red));
+    }
+
+    @ReceiveEvent
+    public void onLocalPlayerInitialized(LocalPlayerInitializedEvent event, EntityRef entity) {
+            Notification notification = new Notification(NOTIFICATION_ID,
+                    "How to open the Statistics Screen?",
+                    "Press " + getActivationKey(new SimpleUri("LightAndShadow:statistics")) + " to access the statistics screen",
```
jdrueckert:
```suggestion
                    "Press " + getActivationKey(new SimpleUri("LightAndShadow:statistics")) + " to see statistics",
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientStatisticsSystem.java
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                })
+                .map(key -> FontColor.getColored(key, Color.yellow))
+                .orElse(FontColor.getColored("n/a", Color.red));
+    }
+
+    @ReceiveEvent
+    public void onLocalPlayerInitialized(LocalPlayerInitializedEvent event, EntityRef entity) {
+            Notification notification = new Notification(NOTIFICATION_ID,
+                    "How to open the Statistics Screen?",
```
jdrueckert: I would definitely go with something shorter. For the other notifications we used something not so "technical" or "official" or "serious", but rather something remotely funny or that had a kind of "ring" to it...

For this I'd propose 
```suggestion
                    "The Numbers Game",
```

jdrueckert (changes requested):
I think it might be better to have to hold it to see statistics rather than to press it and press it again for the screen to vanish.
Thus, players can easily have a really quick peek at the stats during the game without having to deal with the screen obstructing their view too long

assets/ui/statisticsScreen.ui
ahv15: This seems like the easiest way to align the elements (labels) in the necessary layout, I don't think we would have the flexibility in arrangement otherwise.

jdrueckert: I'll need to have another look at the screen in-game, but this seems weird... on the other hand TeraNUI is kind of weird, so maybe it's just another "oh wow, it's _that_ weird" moment for me :sweat_smile:

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientStatisticsSystem.java
```diff
@@ -0,0 +1,117 @@
...
+                })
+                .map(key -> FontColor.getColored(key, Color.yellow))
+                .orElse(FontColor.getColored("n/a", Color.red));
+    }
+
+    @ReceiveEvent
+    public void onLocalPlayerInitialized(LocalPlayerInitializedEvent event, EntityRef entity) {
+            Notification notification = new Notification(NOTIFICATION_ID,
+                    "The Numbers Game",
+                    "Press " + getActivationKey(new SimpleUri("LightAndShadow:statistics")) + " to see statistics screen",
```
jdrueckert: Either go with 
```suggestion
                    "Press " + getActivationKey(new SimpleUri("LightAndShadow:statistics")) + " to see statistics",
```
as I proposed originally, or with
```suggestion
                    "Press " + getActivationKey(new SimpleUri("LightAndShadow:statistics")) + " to see the statistics screen",
```

Currently, it's a bit weird English :sweat_smile: 
I would prefer the first as it's shorter.

src/main/java/org/terasology/ligthandshadow/componentsystem/input/TabButton.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.terasology.engine.input.BindButtonEvent;
+import org.terasology.engine.input.DefaultBinding;
+import org.terasology.engine.input.RegisterBindButton;
+import org.terasology.input.InputType;
+import org.terasology.input.Keyboard;
+
+@RegisterBindButton(id = "statistics", description = "Show statistics screen", category = "general", repeating = true)
+@DefaultBinding(type = InputType.KEY, id = Keyboard.KeyId.V)
+public class TabButton extends BindButtonEvent {
```
jdrueckert: From the name it's not clear that this is an event.
Also I'm missing documentation.
And I wonder whether you mean ["tap"](https://www.oxfordlearnersdictionaries.com/definition/english/tap_1), not ["tab"](https://www.oxfordlearnersdictionaries.com/definition/english/tab_2)? The second explanation for "tab" might fit, but is ambiguous IMO as we're not actually using the tab key...

assets/ui/statisticsScreen.ui
jdrueckert: So _theoretically_, it should be possible to span elements as explained in the old [TeraNUI Tutorial Wiki](https://github.com/Terasology/TutorialNui/wiki/MigLayout#merging-and-splitting-cells). However, I gave it a quick try and it doesn't seem to work anymore... So I guess we need to stick with this for now.

src/main/java/org/terasology/ligthandshadow/componentsystem/input/TabButton.java
ahv15: Most of the button based events I have seen don't explicitly mention event in their name.

skaldarnar: @jdrueckert and I briefly discussed this offline. All the button events (for keys on the keyboard) don't have the suffix, so omitting it for consistency makes sense :+1: 

The only events where we use an `...Event` suffix is for button events associated with mouse clicks - for whatever reason :shrug:

module.txt
```diff
@@ -50,6 +50,11 @@
             "id": "LightAndShadowResources",
             "minVersion": "1.1.0"
         },
+        {
+            "id": "Notifications",
+            "minVersion": "0.1.0",
+            "maxVersion": "0.3.0"
```
skaldarnar, line 56: What's the reason to have a version range here?

overrides/Notifications/ui/NotificationAreaOverlay.ui
```diff
@@ -0,0 +1,28 @@
+{
```
skaldarnar, line 2: I think our JSON assets are parsed in a way that we allow for comments in them.

I'd appreciate if we could add why we need this override for Light and Shadow.

If I get it correctly, this is do move the notification area to the bottom left of the screen (instead of top right corner)?

```suggestion
/** Move the notification area to bottom-left corner of the screen to avoid overlaps with inventory. */
{
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientStatisticsSystem.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @ReceiveEvent
+    public void onLocalPlayerInitialized(LocalPlayerInitializedEvent event, EntityRef entity) {
+            Notification notification = new Notification(NOTIFICATION_ID,
+                    "The Numbers Game",
+                    "Press " + getActivationKey(new SimpleUri("LightAndShadow:statistics")) + " to see statistics",
+                    "engine:items#blueBook");
+            localPlayer.getClientEntity().send(new ShowNotificationEvent(notification));
+    }
```
skaldarnar, line 116: We show this notification as soon as the player joins the game, right? Does it overlap with the dialog box of the fool?

module.txt
ahv15: nothing in particular, was just following the ranges specified when being used in the other modules (Eventual Skills and Ingame Help)

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ClientStatisticsSystem.java
ahv15: Yes looks like it does, the only place that remains is the top left.

module.txt
skaldarnar: Okay, then I guess I have to ask myself why I put them in there that way :thinking: :sweat_smile: 

I guess with the leading `0.` each minor version upgrade is incompatible, and since I knew that 0.1.0 and 0.2.0 are still backwards compatible I put in this range ...

skaldarnar (approved):

jdrueckert (approved):

PR #226, opened 2021-07-08
feat: use item activation (right click) for magic staff

skaldarnar:
This PR makes use of the PRs referenced below to turn the magic staff weapon into an item that is used with "item activation", i.e., right click. As the staff is supposed to be used on range against another player this use case needs to be supported by the engine.

It improves the game play in the sense that using the staff is now similar to using the bow or the fire ball staff instead of a "ranged use" action (as in, opening a door with `e`).

- the staff's effect currently is instant. if we want to nerf it in the future, e.g., by requiring a player to track the enemy flag bearer for some time _t_, how can we do this with the technical foundation we have here? does the activation event help here, or do we need something different?

Depends on MovingBlocks/Terasology#4818
Depends on Terasology/LightAndShadowResources#70

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
@@ -3,6 +3,8 @@
 
 package org.terasology.ligthandshadow.componentsystem.controllers;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.terasology.engine.entitySystem.entity.EntityManager;
 import org.terasology.engine.entitySystem.entity.EntityRef;
 import org.terasology.engine.entitySystem.event.Event;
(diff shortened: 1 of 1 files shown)


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -3,62 +3,78 @@
...
+     */
+    @ReceiveEvent(components = {FlagDropOnActivateComponent.class, ItemComponent.class})
+    public void dropFlagOnRangeAttack(ActivateEvent event, EntityRef item) {
+        if (event.getTarget().exists()
+                && event.getTarget().hasComponent(PlayerCharacterComponent.class)
+                && event.getTarget().hasComponent(HasFlagComponent.class)) {
+
+            logger.info(event.toString());
+            logger.info(item.toFullDescription());
+
```
skaldarnar:
```suggestion
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -3,62 +3,78 @@
...
 import org.terasology.module.inventory.events.InventorySlotChangedEvent;
-import org.terasology.engine.logic.players.PlayerCharacterComponent;
-import org.terasology.ligthandshadow.componentsystem.components.FlagDropOnActivateComponent;
-import org.terasology.ligthandshadow.componentsystem.components.HasFlagComponent;
-import org.terasology.lightandshadowresources.components.RaycastOnActivateComponent;
-import org.terasology.lightandshadowresources.components.LASTeamComponent;
 
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class AttackSystem extends BaseComponentSystem {
+    private static final Logger logger = LoggerFactory.getLogger(AttackSystem.class);
```
skaldarnar:
```suggestion
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -3,62 +3,78 @@
 
 package org.terasology.ligthandshadow.componentsystem.controllers;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
skaldarnar:
```suggestion
```

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -3,62 +3,78 @@
...
-        EntityRef attackingPlayer = event.getInstigator(); // The player using the staff to attack
-        if (canPlayerAttack(attackingPlayer)) {
-            if (targetPlayer.hasComponent(PlayerCharacterComponent.class) && targetPlayer.hasComponent(HasFlagComponent.class)) {
-                // If the target player has the flag
-                targetPlayer.send(new DropFlagEvent(attackingPlayer));
-            }
-        }
-    }
-
-    private boolean canPlayerAttack(EntityRef attackingPlayer) {
```
skaldarnar, line 56: we don't need this anymore as the "raycast" check now happens in the engine when creating the `ActivateEvent`.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/AttackSystem.java
```diff
@@ -3,62 +3,78 @@
...
+    /**
+     * Drop the flag if a player targets the enemy flag bearer with an item that causes a flag drop on activation.
+     * @param event the activation event, must target the enemy flag bearer
+     * @param item the target of the activation (here: the used item)
+     */
+    @ReceiveEvent(components = {FlagDropOnActivateComponent.class, ItemComponent.class})
+    public void dropFlagOnRangeAttack(ActivateEvent event, EntityRef item) {
+        if (event.getTarget().exists()
+                && event.getTarget().hasComponent(PlayerCharacterComponent.class)
+                && event.getTarget().hasComponent(HasFlagComponent.class)) {
```
skaldarnar, line 55: This is also checked in `dropFlagOnPlayerAttack`, so we can probably drop this part of the check here... :thinking:

ahv15 (dismissed):
The changes work pretty well and it doesn't cause the problems I initially thought it would cause.

ahv15:
It might be a good idea to shift the staff to the shop (once it's added) instead as now players can remove the flag even without the staff and having a larger range (the staff) shouldn't be the default behavior (as all the players could camp and prevent others from bringing the flag back) and making sure that the player tracks the opponent would probably be necessary (not sure whether occasionally checking whether the target is being detected during the time period would be the right thing to do) as the weapon might be too powerful based on the range.

jdrueckert (approved):

PR #228, opened 2021-07-10
feat: implementation for economy

ahv15:
Makes use of the implementation from GooeyDefence and requires https://github.com/Terasology/Economy/pull/22, and needs https://github.com/Terasology/LightAndShadowResources/pull/72 to test the shop screen.


assets/skins/DefenceDefault.skin
```diff
@@ -0,0 +1,29 @@
+{
```
skaldarnar: this should have different name

skaldarnar: if we believe this is a good default for ... shop screens ... we should just add the skin to Economy.

module.txt
```diff
@@ -34,6 +34,10 @@
             "id": "Dialogs",
             "minVersion": "1.0.0"
         },
+        {
+            "id": "Economy",
+            "minVersion": "1.1.0-SNAPSHOT"
```
jdrueckert: As @skaldarnar commented in the associated GD PR: we might want to release `Economy` once the PR is merged and bump the version number.

overrides/Inventory/ui/inventoryScreen.ui
```diff
@@ -0,0 +1,219 @@
+{
+  "type": "ShopScreen",
```
jdrueckert: I wonder if this can live in `Economy` as well? Or does it differ from the one we have in GD? If so, in how far?

ahv15:
This should fix the multiplayer issues.

jdrueckert (approved):

PR #229, opened 2021-07-17
feat: face opponents base on spawn

ahv15:
Requires https://github.com/MovingBlocks/Terasology/pull/4823 to test this.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
+++ b/src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
@@ -178,6 +178,14 @@ public final class LASUtils {
         return null;
     }
 
+    public static float getYaw(String team) {
+        return (float) Math.toDegrees(Math.atan2(LASUtils.getTeleportDestination(team).
+                sub(LASUtils.getTeleportDestination(getOppositionTeam(team)), new Vector3f()).
+                cross(new Vector3f(0, 0, 1)).dot(0, 1, 0), LASUtils.getTeleportDestination(team).
+                sub(LASUtils.getTeleportDestination(getOppositionTeam(team)), new Vector3f()).
+                dot(new Vector3f(0, 0, 1))));
+    }
+
     public static boolean circle(
             int x,
             int y,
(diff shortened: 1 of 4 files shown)


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -65,6 +66,8 @@ public void beforeDestroy(BeforeDestroyEvent event, EntityRef player, CharacterC
             player.send(new RestoreFullHealthEvent(player));
             Vector3f randomVector = new Vector3f(-1 + random.nextInt(3), 0, -1 + random.nextInt(3));
             player.send(new CharacterTeleportEvent(randomVector.add(LASUtils.getTeleportDestination(team))));
+            float playerYaw = team.equals(LASUtils.RED_TEAM) ? -90 : 90;
```
skaldarnar: While this works currently for our simple map, this code would be more resilient if we'd compute the yaw from the teleport target position and the opponent player base position.

skaldarnar (commented):
The fact that we have to change code in three places worries me a bit. We should probably take a look at how we can refactor this.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
ahv15: Made use of this https://stackoverflow.com/questions/5188561/signed-angle-between-two-3d-vectors-with-same-origin-within-the-same-plane as the default calculation of angles isn't signed. The 2nd comment gave me a more precise answer so I used that.

src/main/java/org/terasology/ligthandshadow/componentsystem/LASUtils.java
```diff
@@ -178,6 +178,14 @@ public static String getOppositionTeam(String team) {
         return null;
     }
 
+    public static float getYaw(String team) {
```
jdrueckert: `getYaw` seems a good name for a general utility function, but doesn't match the hard-coded (and not documented) "assumptions" about _what_ yaw we're calculating here (relation to the teleport destinations of the teams, etc.)

skaldarnar (approved):

skaldarnar:
Builds are failing because of https://github.com/Terasology/LightAndShadowResources/pull/70#issuecomment-883678945

PR #230, opened 2021-07-19
feat: use cards for defense

ahv15:
Get a card using ```give clubsAce```. Point it at the ground and activate it to spawn a door like structure which can be used as a passive defense mechanism. The item itself still needs some tweaks,
image

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
@@ -1,31 +1,15 @@
-/*
- * Copyright 2017 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
 
 package org.terasology.ligthandshadow.logic;
 
 import org.joml.Vector3f;
 import org.joml.Vector3i;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.terasology.engine.audio.AudioManager;
 import org.terasology.engine.entitySystem.entity.EntityBuilder;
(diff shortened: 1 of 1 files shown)


src/main/java/org/terasology/ligthandshadow/logic/CardSystem.java
```diff
@@ -89,31 +64,30 @@ public void placeCard(ActivateEvent event, EntityRef entity) {
...
+        for (String s : primeBlock.getBlockFamily().getCategories()) {
+            if (s.equals("soil")) {
+                exit = false;
+                break;
+            }
+        }
+        if (exit) {
             event.consume();
             return;
         }
```
jdrueckert: Why this check for soil?

ahv15: This is so that the card is only placed above the ground (soil) rather than on top of grass, flowers, the bases etc.

jdrueckert: in that case, I'd rather go for an attribute such as "solid"... I think you could use the block property check `isPenetrable` here 🤔 (see https://github.com/MovingBlocks/Terasology/blob/develop/engine/src/main/java/org/terasology/engine/world/block/Block.java#L230)

ahv15: I don't think that check would be sufficient for more specific scenario's such as when a player tries to place the card on the base and the floating platform. Not sure but I will test it out later.

jdrueckert: do we want to forbid to place anything on the base? consider that you would also need to check something like that for everything that is placeable then.
on the floating platform, a player without cheat permissions should not be able to get a card in the first place 🤔

ahv15: the shop is available even on the floating platform. By placing the cards on the base it becomes possible to barricade the flag itself which might make it too difficult to get the flag (while breaking down the barrier the player's defenses become too low).

jdrueckert: > the shop is available even on the floating platform

that's something I'm not entirely convinced we want (I mean that's what we have the pre-game phase for, right?) to be honest...

> it becomes possible to barricade the flag itself

I'd say this is exactly the kind of defense strategy that the cards could be used for and I don't see too much harm in that in general.
In the end it's a question of balancing (which is not reasonable to do at this point in time - yet). If in comparison to other mechanisms the cards become too powerful, we can still tweak that.
Restricting the placement is one screw you can turn, but there's also others, i.e. card price, max number of cards you can buy, etc ... But IMO restricting that already now would be premature optimization.

jdrueckert (approved):
Tests out fine and looks good code-wise. Any improvements to the item can be done as a follow-up IMO.

PR #231, opened 2021-08-03
feat: damage multiBlock Structures

ahv15:
Whenever a block in a multiBlock structure is damaged all the others blocks in the multiBlock structure are also damaged.


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/MultiBlockSystem.java
```diff
@@ -0,0 +1,57 @@
...
+            damage = item.baseDamage;
+            if (item.damageType != null) {
+                damageType = item.damageType;
+            }
+        }
+        if (entity.hasComponent(MultiBlockMainComponent.class)) {
+            blocks = entity.getComponent(MultiBlockMainComponent.class).getMultiBlockMembers();
+        } else if (entity.hasComponent(MultiBlockMemberComponent.class)) {
+            blocks = blockEntityRegistry.getBlockEntityAt(entity.getComponent(MultiBlockMemberComponent.class).getMainBlockLocation()).getComponent(MultiBlockMainComponent.class).getMultiBlockMembers();
+        }
```
skaldarnar: I'd do this in two separate steps:

1. find the main multi-block, if any
2. derive the `blocks` from that main block

To make the code easier to comprehend I'd additionally move this to a method which returns an `Optional<Collection<Vector3i>>` with the multi-block member positions, if any multi-block is hit.

As it does not make any difference for us whether we iterate over an empty collection (and thus don't damage any other blocks) or check if the collection is present we may even get rid of the optional here and just return a collection that may or may not contain any entries.

skaldarnar (changes requested):
I'm surprised this is not a default feature of MultiBlocks :eyes:  What does MB do with attacks to a multi-block structure?

PR #232, opened 2021-08-10
refactor: clean up Score system

ahv15:


src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/ScoreSystem.java
```diff
@@ -97,12 +97,7 @@ private void checkAndResetGameOnScore(ActivateEvent event, EntityRef entity) {
     }
 
     private void sendGameOverEventToClients(String winningTeam) {
-        if (entityManager.getCountOfEntitiesWith(ClientComponent.class) != 0) {
-            Iterable<EntityRef> clients = entityManager.getEntitiesWith(ClientComponent.class);
-            for (EntityRef client : clients) {
-                client.send(new GameOverEvent(winningTeam, blackScore, redScore));
-            }
-        }
+        sendEventToClients(new GameOverEvent(winningTeam, blackScore, redScore));
```
jdrueckert: with this, I think we even can drop this method and inline this in `checkAndResetGameOnScore`

jdrueckert: other than that, I think this should be fine... @skaldarnar do you see anything you'd like cleaned up some more?

skaldarnar (approved):

PR #233, opened 2021-08-15
fix: Destroy magic barriers and make players vulnerable on headless servers

ahv15:
There does seem to be a 1-2 seconds delay between the UI and the destruction of the barrier. Not sure how to test this on a headless server.

### DIFF
--- a/src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
+++ b/src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
@@ -24,6 +24,7 @@ import org.terasology.ligthandshadow.componentsystem.events.DelayedDeactivateBar
 
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class MagicDomeSystem extends BaseComponentSystem {
+    private static final String DEACTIVATE_BARRIERS_ACTION = "LightAndShadow:deactivateBarriers";
     private static final int PREGAME_ZONE_RADIUS = 20;
     @In
     private EntityManager entityManager;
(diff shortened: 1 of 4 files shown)


src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
```diff
@@ -47,9 +51,16 @@ public void activateBarriers(ActivateBarrierEvent event, EntityRef entity) {
     }
 
     @ReceiveEvent
-    public void deactivateBarriers(DeactivateBarrierEvent event, EntityRef entity) {
-        redBarrier.destroy();
-        blackBarrier.destroy();
+    public void delayedDeactivateBarriers(DelayedDeactivateBarrierEvent event, EntityRef entity) {
+        delayManager.addDelayedAction(entity, "deactivate", 30000);
```
skaldarnar: A) I think the time after which the barriers should be destroyed should be defined by the event.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/TeleporterSystem.java
```diff
@@ -103,6 +104,7 @@ private boolean isProperTeamSize(EntityRef teleporter, EntityRef player) {
         if (teleporterTeamCount - oppositeTeamCount < maxTeamSizeDifference) {
             if (teleporterTeamCount >= 0 && oppositeTeamCount >= 1 && !gameStart) {
                 sendEventToClients(TimerEvent::new);
+                player.send(new DelayedDeactivateBarrierEvent());
```
skaldarnar: If we call this a "delayed" event, we should encode the delay in this event, so that we only have a single source of truth how long the countdown is.

src/main/java/org/terasology/ligthandshadow/barrier/MagicDomeSystem.java
skaldarnar: B) These delayed action ids are global, and "deactivate" is pretty generic. I'd recommend to define a constant with a name resembling our typical asset urn schema, e.g., 

```java
private static final String DEACTIVATE_BARRIERS_ACTION = "LightAndShadow:deactivateBarriers";
```

This constant should be used both here and in the event handler for the delayed action

skaldarnar (changes requested):

skaldarnar (approved):

PR #239, opened 2021-10-08
Journal integrated with gaming instruction's screenshots

ujjman:
I have integrated Journal module with LAS to show gaming instructions with the help of screenshots

To test the PR, start game with LightAndShadow Tempalate and then interact with MagicFool (by pressing E) then you will get the Journal (check it by pressing J)


src/main/java/org/terasology/journal/integration/LASJournalIntegration.java
```diff
@@ -0,0 +1,69 @@
...
+        Prefab screenshot3 = prefabManager.getPrefab("LightAndShadow:LASTutorial Screenshot 3");
+        Prefab screenshot4 = prefabManager.getPrefab("LightAndShadow:LASTutorial Screenshot 4");
+        chapterHandler.registerJournalEntry("Instructions",
+                Arrays.asList(
+                        new ScreenshotParagraph(new Prefab[]{screenshot1, screenshot2, screenshot3, screenshot4}, null)
+                ));
+        journalManager.registerJournalChapter(lasChapterId,
+                Assets.getTextureRegion("LightAndShadow:MagicFoolIcon").get(),
+                "LightAndShadow", chapterHandler);
+        logger.info("registered journal chapter");
```
jdrueckert, line 62: This should be removed. _If_ you want to introduce logs for the sake of debugging for instance, please use `debug` level.

jdrueckert (commented):
Again here, I don't think we should make use of `Item:icon`.
Style-wise note: filenames should not have whitespaces and ideally try to name things more clearly, so instead of numbering them "Screenshot 1..4", rather name them according to what they display.

Issue #234, opened 2021-08-29
Crash in PlayerDeathSystem while trying to update statistics


skaldarnar:
During play-testing https://github.com/MovingBlocks/Terasology/releases/tag/v5.2.0-rc.1 the (headless) server crashed due to the following NPE when updating the player statistics:

```
java.lang.NullPointerException: null
        at org.terasology.ligthandshadow.componentsystem.controllers.PlayerDeathSystem.updateStatistics(PlayerDeathSystem.java:93)
        at org.terasology.ligthandshadow.componentsystem.controllers.PlayerDeathSystem.beforeDestroy(PlayerDeathSystem.java:63)
        at org.terasology.ligthandshadow.componentsystem.controllers.PlayerDeathSystemMethodAccess.invoke(Unknown Source)
        at org.terasology.engine.entitySystem.event.internal.EventSystemImpl$ByteCodeEventHandlerInfo.invoke(EventSystemImpl.java:379)
        at org.terasology.engine.entitySystem.event.internal.EventSystemImpl.sendConsumableEvent(EventSystemImpl.java:254)
        at org.terasology.engine.entitySystem.event.internal.EventSystemImpl.send(EventSystemImpl.java:233)
        at org.terasology.engine.core.bootstrap.eventSystem.AbstractEventSystemDecorator.send(AbstractEventSystemDecorator.java:67
```

This occurs when the player does not have a `PlayerStatisticsComponent`.

https://github.com/Terasology/LightAndShadow/blob/8c937cdc727b3e342523c45020c3ffc66eb2c765/src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java#L90-L99

jdrueckert:
![](https://github.trello.services/images/mini-trello-icon.png) [Release v5.2.0](https://trello.com/c/iI2iaodH/181-release-v520)

jdrueckert:
Fixed by #242

PR #242, opened 2021-10-26
fix: don't update kill statistics if player died of own account

ujjman:
This PR is to fix issue https://github.com/Terasology/LightAndShadow/issues/234


DarkWeird (dismissed):
Looks right.
event.investigator is NullRef when damage happens from falling.

jdrueckert (changes requested):
@ujjman I would argue that if the player doesn't have the `PlayerStatisticsComponent` yet and we want to update statistics, that instead of returning without doing anything, we should create the component :thinking: or am I overlooking anything here?

ujjman:
I think we can update the deaths counter of the player but I think it would be unfair to do so because the player died by itself and not being killed by the other player. So what do you think?

jdrueckert:
That the player wasn't killed by a person but on their own account is a good point, but it's still a death, so is it really unfair? 🤔 I would be interested in what @skaldarnar has to say about this as we talked about not punishing players, but only rewarding them...

jdrueckert:
@ujjman Had a quick check offline with @skaldarnar and he agrees, that we should count it in the deaths statistic.

src/main/java/org/terasology/ligthandshadow/componentsystem/controllers/PlayerDeathSystem.java
```diff
@@ -90,6 +90,9 @@ private void dropItemsFromInventory(EntityRef player) {
 
     private void updateStatistics(EntityRef player, String type) {
         PlayerStatisticsComponent playerStatisticsComponent = player.getComponent(PlayerStatisticsComponent.class);
+        if (player.getId() == 0) {
```
jdrueckert: Question 1: When is the player ID equal to 0 and why do we want to simply return in that case?
Question 2: What happens if a player a different ID dies on their own account but doesn't have a `PlayerStatisticsComponent` yet?

ujjman: let me give a brief explanation of what I have done. So, when the player dies by itself, `beforeDestroy()` calls `updateStatistics()`. 

The first time `updateStatistics()` is called with parameter `event.instigator` (which will be NullRef in this case as player dies by itself, returning with EntityRef of ID=0) and `"kills"`. So, in this case we cannot update statistics.

The Second time when `updateStatistics()` is called with parameter `player` and `"deaths"`, we get `EntityRef` of the player (initially having `PlayerStatisticsComponent`). So, in this case we update the player's deaths.

Answer 1: Player ID=0 when `event.instigator` is called and player dies by itself. We return in that case because we cannot update the statistics of NullRef (also no point of updating any stats here)

Answer 2: The player gets `PlayerStatisticsComponent` as soon as the player gets spawned on the fool's platform

jdrueckert: Ohh... so basically you're saying that we run into this NPE because in some cases `updateStatistics` is called with a `NullRef` instead of a player entity and obviously `NullRef` won't have the component we're trying to update?
Interesting find!

I'm wondering, whether we want to fix this here though 🤔 Ideally `updateStatistics` should always be called with a player entity and it also doesn't really know anything about the possible causes for being called with a `NullRef` (e.g. player dying without any other entity causing it). So I'd say we'd rather want to fix this in `beforeDestroy` and only call `updateStatistics` for "kills" if the event instigator of the `BeforeDestroyEvent` is not `NullRef`.

And while we're talking about this: For the spear I think this means that when the spear is thrown, we forgot who threw it by the time it hits and kills a player - hence the event instigator is `NullRef` and we run into the NPE 💡
With this PR fixing the NPE, I think we will see that killing a player by throwing a spear will increase their death count but not our kill count, so  this is a separate bug we need to document (as a GitHub issue) and fix. Can you please create a ticket for this and (in a separate PR) investigate why we forget who threw it and how we can memorize that information?

ujjman: ok, will investigate on this issue  👍

jdrueckert (approved):

PR #244, opened 2021-11-07
fix: repeated pregame phase

jdrueckert:


jdrueckert:
Figured it'd be better to do a thorough design of the state machine for LaS instead of monkey-patching, so I'll close this for now and start working on doing this properly instead.

PR #245, opened 2021-11-10
refactor: merge las and ligthandshadow pkgs into lightandshadow pkg

jdrueckert:
This PR is superseding https://github.com/Terasology/LightAndShadow/pull/209 which has gone out of sync due to the gestalt component migration. It does a less fine-granular re-ordering of the files and simply aims to merge the two legacy packages by moving the files into the following structure:
- components
- events
- systems
- ui
- world

In the context of the state-machine-based rework of this module, the structure is bound to change, so I don't think it makes too much sense to invest a lot of time into a fine-grained file structure yet.

Supersedes #209


jdrueckert:
[LaS Packages Merge](https://trello.com/c/RrccK2U8/196-las-packages-merge)

skaldarnar (approved):
I don't see any blocker for this. Light and Shadow is a top-level gameplay module that is not used by other modules affected by this package restructuring 👍 

Also big thanks to @ahv15 for the first attempt!

PR #250, opened 2021-11-21
feat: making players invulnerable before game round starts

ujjman:
This PR adds `InvulnerableComponent` to the players before the match to prevent the players from accidentally dying or killing each other before the actual game rounds start

### DIFF
--- a/deltas/engine/prefabs/player/player.prefab
+++ b/deltas/engine/prefabs/player/player.prefab
@@ -1,10 +1,11 @@
 {
-    "Character": {
-        "nameTagOffset": 1.7
-    },
-    "LASTeam" : {
-        "team" : "white"
-    },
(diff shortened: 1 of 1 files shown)


jdrueckert (commented):
Please note, that this PR is not actually _fixing_ the NPE. It only results in the _side effect_ that players won't die and, thus, won't run into the NPE situation. This is an important difference.

So please adjust the PR title and description to reflect that this PR is about making players invulnerable before the game actually starts to avoid players killing each other or accidentally dying before the start of an LaS round.

jdrueckert (commented):
Approved "code"-wise, but this still needs testing which unfortunately I cannot do at the moment.

jdrueckert (approved):
Tests out fine :+1:

PR #252, opened 2021-11-30
fix: NPE while using kill command on platform

ujjman:
This PR will fix the NPE that occurs while using `kill` command on the fool's platform and teleports the player back to initial spawning position

### DIFF
--- a/src/main/java/org/terasology/module/lightandshadow/LASUtils.java
+++ b/src/main/java/org/terasology/module/lightandshadow/LASUtils.java
@@ -63,6 +63,7 @@ public final class LASUtils {
     // The position near the team's base that player will be teleported to on choosing a team
     public static final Vector3fc RED_TELEPORT_DESTINATION = new Vector3f(29, 12, 0);
     public static final Vector3fc BLACK_TELEPORT_DESTINATION = new Vector3f(-29, 12, 0);
+    public static final Vector3fc WHITE_TELEPORT_DESTINATION = new Vector3f(0, 61, 252);
     public static final String MAGIC_STAFF_URI = "LightAndShadowResources:magicStaff";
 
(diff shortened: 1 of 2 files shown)


jdrueckert (commented):
This does not yet work as expected. When using `/kill` this will kill and port the player that _initiates_ the command instead of the one _targeted_ by the command :thinking: 
To make sure it's not unrelated to this PR, I'll test in a sec, whether this also happens with current develop.

jdrueckert:
Oh nevermind, I just noticed, that the command is _expected_ to kill the player using it. It doesn't have any parameter that would allow to specify which player to kill. Might be a neat tooling addition, though, to add a kill command for administrators that allows killing a specific player. Would help if they get stuck for whatever reason, for instance.

src/main/java/org/terasology/module/lightandshadow/LASUtils.java
```diff
@@ -162,6 +166,9 @@ public static String getOppositionTeam(String team) {
         if (team.equals(BLACK_TEAM)) {
             return RED_TEAM;
         }
+        if (team.equals(WHITE_TEAM)) {
+            return WHITE_TEAM;
+        }
```
jdrueckert, line 171: What do we need this for?

ujjman: this is just to prevent NPE that can occur on https://github.com/Terasology/LightAndShadow/blob/develop/src/main/java/org/terasology/module/lightandshadow/systems/PlayerDeathSystem.java#L71

ujjman: also this will prevent the player from facing to the direction of red/black team and will make the player to face magic fool

jdrueckert (approved):

PR #254, opened 2021-12-04
fix: prevent npe on player death by thrown spear

skaldarnar:
Combat system may assign `null` as instigator to some (Before)DestroyEvents which causes an NPE in the player death system.

This fixes the crash with a `TODO` item to find and fix the root cause. See https://github.com/Terasology/CombatSystem/pull/92 for first investigation results.

In addition, this commit introduces an enum for the statistics type to update to avoid any bugs resulting from using plain string "constants" (we did not even use constants ...).

Furthermore, I've made the pre-game countdown a constant. This should make it easier to find this value and adjust it later, e.g., for local testing or tweaking.

### DIFF
--- a/src/main/java/org/terasology/module/lightandshadow/systems/ClientPregameSystem.java
+++ b/src/main/java/org/terasology/module/lightandshadow/systems/ClientPregameSystem.java
@@ -26,6 +26,8 @@ public class ClientPregameSystem extends BaseComponentSystem {
 
     private static final String PREGAME_MESSAGE = "The game start's as soon as there is \n at least one player in each team.";
 
+    private static final int COUNTDOWN_IN_SECONDS = 30;
+
     private static Timer timer;
 
     @In
@@ -60,7 +62,7 @@ public class ClientPregameSystem extends BaseComponentSystem {
         if (localPlayer.getClientEntity().equals(entity)) {
             timer = new Timer();
             timer.scheduleAtFixedRate(new TimerTask() {
-                int timePeriod = 30;
+                int timePeriod = COUNTDOWN_IN_SECONDS;
                 boolean addNotification;
                 public void run() {
                     if (addNotification) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/terasology/module/lightandshadow/systems/ClientPregameSystem.java
```diff
@@ -26,6 +26,8 @@
 
     private static final String PREGAME_MESSAGE = "The game start's as soon as there is \n at least one player in each team.";
 
+    private static final int COUNTDOWN_IN_SECONDS = 30;
+
```
jdrueckert, line 47: Should this reside with the other LAS constants in LASUtils?

src/main/java/org/terasology/module/lightandshadow/systems/PlayerDeathSystem.java
```diff
@@ -43,17 +43,17 @@
...
      * This is a high priority method, hence it receives the event first and consumes it.
      * This prevents the destruction of player entity and prevents the deathScreen from showing up.
      *
-     * @param event
-     * @param player
-     * @param characterComponent
-     * @param aliveCharacterComponent
+     * @param event notification event that a player entity is about to be destroyed (usually sent on death)
+     * @param player the player entity about to be destroyed
+     * @param characterComponent ensures that the entity has a character (TODO: why do we need this?)
```
jdrueckert, line 55: To make sure we only fetch this event for actual players? :thinking: 
As we only do something here if `PlayerCharacterComponent` is present, maybe we want to "require" that one instead.

src/main/java/org/terasology/module/lightandshadow/systems/PlayerDeathSystem.java
```diff
@@ -43,17 +43,17 @@
...
+     * @param aliveCharacterComponent ensures that the player is currently alive (TODO: parameter not used, move to annotion?)
```
jdrueckert:
```suggestion
     * @param aliveCharacterComponent ensures that the player is currently alive (TODO: parameter not used, move to annotation?)
```

jdrueckert (dismissed):

jdrueckert (approved):

PR #256, opened 2021-12-17
feat: implement state machine to model game state

skaldarnar:


jdrueckert:
Currently it seems that `OnPreGamePhaseStartedEvent`, albeit being a `BroadcastEvent`, isn't broadcasted to all clients. At least when I debug in, I only run into a breakpoint at https://github.com/Terasology/LightAndShadow/blob/67a0ea98726f2b50999857c0e49853291c6c2ccc/src/main/java/org/terasology/module/lightandshadow/systems/ClientPregameSystem.java#L78 once (for the hosting client) even if I have two clients... no idea yet why that is.

